PR #24, opened 2021-03-08
Component lifecycle

### LINKED ISSUE #21: Component lifecycle
Components should have some lifecycle hooks for `on_created` and `on_destroyed`.

`on_created` is similar to `create_effect` except it only runs once.
`on_destoryed` is a bit more complicated. Currently, components are simply function calls. For `on_destroyed` to be implemented, the function call needs to be wrapped up in another `create_component` utility. This utility function will setup the lifecycle hooks for the code inside the component.

**Additional context**

**Questionnaire**
- [x] I would like to implement this feature and I have a solution

lukechu10:
Fixes #21 

Adds support for the `on_cleanup` hook. `on_cleanup` is tied to the reactive scope. When the reactive scope (aka `Owner`) is dropped, all the cleanup callbacks are called. Components are created in a new reactive scope so calling `on_cleanup` inside a component will trigger the callback when the component is destroyed.

### DIFF
--- a/maple-core/src/lib.rs
+++ b/maple-core/src/lib.rs
@@ -75,7 +75,7 @@ pub mod prelude {
     pub use crate::cloned;
     pub use crate::reactive::{
         create_effect, create_effect_initial, create_memo, create_root, create_selector,
-        create_selector_with, Signal, StateHandle,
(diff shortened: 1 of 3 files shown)


lukechu10:
Blocked by #37

PR #29, opened 2021-03-09
Refactor effect dependency tracking

### LINKED ISSUE #27: Support nested effects
Nested effects are required to implement #18 

Nested effects should be supported via the `create_effect` function.

**Additional context**
Semantics:
- Inner effect has dependencies on state used inside inner closure.
- Outer effect has dependencies on state used inside outer closure but not within inner closure.

Proposed implementation: https://github.com/lukechu10/maple/issues/18#issuecomment-793284196

**Questionnaire**
- [ ] I would like to implement this feature and I have a solution

### LINKED ISSUE #28: Recreate dependency graph on every call in create_effect
**Describe the bug**
The dependency graph should be recreated every time the closure in `create_effect` is called. If there is control flow in the closure or not every expression is executed, not all the states could get tracked.

**Additional context**
Useful blog post: https://dev.to/ryansolid/building-a-reactive-library-from-scratch-1i0p

**Questionnaire**
- [ ] I would like to fix this bug but I don't know where to start
- [x] I would like to fix this bug and I have a solution

lukechu10:
Fixes #27 and #28

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -8,6 +8,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  PUBLIC_URL_PREFIX: "/maple"
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


lukechu10:
This PR breaks basic cycle detection. Cycle detection was not completely working anyways and will be tricky to implement with dynamic dependencies.

PR #31, opened 2021-03-09
Remove double-boxing of Computation

Kestrer:
`Computation` previously had two boxes (or well box-likes) as it was essentially an `Rc<Box<dyn Fn()>>` - so I replaced this with just an `Rc<dyn Fn()>`. Additionally, I switched from a `Vec<Rc<Computation>>` to a `HashSet<Computation>`, which should allow `observe` to become a constant-time instead of linear-time operation.


lukechu10 (approved):
LGTM

PR #42, opened 2021-03-13
Use Clippy

Kestrer:
Clippy is a useful tool for preventing common mistakes. One mistake it caught in the existing code is that we were comparing `*const dyn Fn`s. This compares the vtable pointer as well as the data pointer, but comparing the vtable pointer is entirely meaningless and could cause bugs, so I changed it to just compare the data pointer.


lukechu10:
Looks good!

PR #67, opened 2021-03-27
Backend Abstraction

### LINKED ISSUE #66: Introduce a new abstraction layer in `TemplateResult` for multiple backends
Blocker for #47 

The current `TemplateResult` struct is basically a new-type around `web_sys::Node`. To support SSR, `template!`s should be rendered to a `String` rather than DOM nodes. The `TemplateResult` struct should therefore represent the complete DOM hierarchy with a `.to_node()` method to convert this layer into DOM nodes. This might also be related to #19.

For supporting SSR/pre-rendered applications, `TemplateResult` should have a `.to_string()`.

Note that this is not a VDOM because this abstraction layer is only a wrapper around pure DOM nodes. Updates will still be made directly without any patching or reconciliation phase.

**Describe alternatives you've considered**
Render the DOM in a wasm runtime (e.g. NodeJS, wasmer, etc...) with mocked DOM apis (e.g. [`jsdom`](https://www.npmjs.com/package/jsdom). This is sounds more like a workaround rather than a real solution. We will probably run into future problems with this as well.

**Additional context**
Percy supports SSR: https://chinedufn.github.io/percy/views/server-side-rendering/how-to-ssr.html
Solid-SSR is probably also worth looking at as this library is heavily inspired by SolidJS: https://github.com/ryansolid/solid/tree/master/packages/solid-ssr

**Questionnaire**
- [ ] I would like to implement this feature and I have a solution

lights0123:
Fixes #66. I'm leaving for a week in 2 days, but I just spent an hour or two working on this today, and I'll probably do a little more tomorrow. Feel free to build on this or just do something else completely.

This introduces a new trait, `GenericNode`. Right now, it only has one implementation, which is a real DOM node. I'm writing this in a way so that users will have to make their components generic, but will hopefully never have to specify what that generic type is because it'll be inferred from calling `render()`.

Right now, I have the following questions I'll need to answer to go further:
- Handling `next_sibling` and `parent_node` functions. When those just call JS functions, it's no problem because it has a full GC that can handle cyclic references. Although that's possible in Rust with some work, it would be more efficient if we could remove calls to those functions. I could also make those return an error server-side (they're only used for modifying the state), but it would make modifying the data impossible server-side—meaning you can't have some `async` computation render stuff later, or if you did, it'd have to be implemented differently.
- `NodeRef`: How to do this in a user-friendly way: do we force the user to downcast to a concrete type? It would be nice to not hardcode it being a `web_sys::Node`, for use cases like React Native. However, there is probably no need for it to work on the server.


lights0123:
The basic Hello World demo compiles and works now.

> Handling next_sibling and parent_node functions. When those just call JS functions, it's no problem because it has a full GC that can handle cyclic references. Although that's possible in Rust with some work, it would be more efficient if we could remove calls to those functions. I could also make those return an error server-side (they're only used for modifying the state), but it would make modifying the data impossible server-side—meaning you can't have some async computation render stuff later, or if you did, it'd have to be implemented differently.

I decided just to implement them for now. I'll probably make them error out on the server-side.

lights0123:
Hello World, Counter, Components, and the docs all work now. Todomvc doesn't, but that's only because it uses `NodeRef`. Tomorrow, I'll figure out what to do about that.

lights0123:
Although, before I do that, does this look good? The alternative way would be with feature flags. That may make `NodeRef` usage cleaner, but would be less flexible—if, for whatever reason, you'd like the client to be able to render both to the DOM and a string, you'd be prevented from doing so.

lukechu10 (commented):
Looks good. I think it would make more sense with feature flags because everything that is `wasm-bindgen` or `web-sys` would need to be put behind a conditional compilation flag anyways. Otherwise it won't compile on non `wasm32` targets.
It would also be nice if we didn't have to make everything generic over `G` as that is a bit verbose.

The way you implemented this wasn't exactly the way I thought of doing it but that's OK. I was thinking more along the lines of `GenericNode` having a `.to_node()` method which would generate `DomNode`s from the `GenericNode`. What do you think?

maple-core/src/generic_node/vdom.rs
```diff
@@ -0,0 +1,180 @@
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::fmt;
+use std::rc::Rc;
+
+use crate::generic_node::{EventListener, GenericNode};
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Node {
```
lukechu10, line 9: Continuing from my previous review, how about making `template!` render to a `Node` and use `Node` to generate real DOM. There isn't any VDOM overhead as it is only used on creation.

lukechu10 (commented):
Sorry I missed some of the new changes in my previous review

lights0123:
> everything that is wasm-bindgen or web-sys would need to be put behind a conditional compilation flag anyways. Otherwise it won't compile on non wasm32 targets.

That's not true—it just panics at runtime.

lukechu10:
> That's not true—it just panics at runtime.

Yup you're right. My bad. Although I still don't like the generic type params on all the components. Maybe something could be done using dynamic dispatch?

lukechu10:
Just looked at the `ssr` example you added and it really is quite impressive! Thanks a lot for all the work so far.

lukechu10:
About feature flags, it might be a good idea to add a `ssr` feature anyways to prevent penalizing users who don't use SSR with worse compile times.

maple-core/src/flow.rs
```diff
@@ -281,30 +258,23 @@ where
...
 /// ```
-pub fn Indexed<T, F: 'static>(props: IndexedProps<T, F>) -> TemplateResult
+pub fn Indexed<T, F: 'static, G: GenericNode>(props: IndexedProps<T, F, G>) -> TemplateResult<G>
 where
     T: Clone + PartialEq,
-    F: Fn(T) -> TemplateResult,
+    F: Fn(T) -> TemplateResult<G>,
 {
-    let templates: Rc<RefCell<Vec<(Owner, TemplateResult)>>> = Rc::new(RefCell::new(Vec::new()));
+    let templates: Rc<RefCell<Vec<(Owner, TemplateResult<G>)>>> = Rc::new(RefCell::new(Vec::new()));
```
lights0123, line 268:
```suggestion
    let templates: Rc<RefCell<Vec<(Owner, TemplateResult<G>)>>> = Default::default();
```

maple-core/src/flow.rs
```diff
@@ -66,25 +64,17 @@ where
     let iterable = Rc::new(iterable);
     let key_fn = Rc::new(key_fn);
 
-    type TemplateValue<T> = (Owner, T, TemplateResult, usize /* index */);
+    type TemplateValue<T, G> = (Owner, T, TemplateResult<G>, usize /* index */);
 
     // A tuple with a value of type `T` and the `TemplateResult` produces by calling `props.template` with the first value.
-    let templates: Rc<RefCell<HashMap<Key, TemplateValue<T>>>> =
+    let templates: Rc<RefCell<HashMap<Key, TemplateValue<T, G>>>> =
         Rc::new(RefCell::new(HashMap::new()));
```
lights0123, line 73:
```suggestion
        Default::default();
```

lukechu10:
I'll fix the integration tests and I'll probably merge this. Thanks for all the work @lights0123!

Issue #88, opened 2021-04-06
Add more easing functions


lukechu10:
**Is your feature request related to a problem? Please describe.**
Maple ships with some easing functions out of the box for use with `Tweened`. They are in the `maple_core::easing` module. There are some easing functions that are missing, notably, `circular`, `sine`, `exponential`, `bounce` and perhaps some more.

**Describe the solution you'd like**
For all the different easing functions listed above, there should be `in`, `out` and `inout` variants.

**Additional context**
https://easings.net/ would be a good reference for the right mathematical functions for each ease function.

**Questionnaire**
- [ ] I would like to implement this feature but I don't know where to start
- [ ] I would like to implement this feature and I have a solution
- [ ] I don't have time to implement this / I don't want to implement this

tmpr:
Hello! I would like to implement the `circular` and `exponential` functions! This is my first time ever contributing to some other repository, but I will give my best!

I do have a plan, yet not a solution.

PR #90, opened 2021-04-06
Add specified easing functions specified in (#88)

tmpr:
### What was added

Functions for circular, sine, exponential and bouncing easing have been added to `easing.rs`. Further, I added 3 constants. This doesn't have to close (#88), since one could still add functions.

### Remarks

- I took most functions from [https://easings.net/](here), yet had no idea yet how to test the functions.
- A few things about the bouncing easing could be added improved:
     -  One could clarify the choice of numbers
     - Some of the code could be more compact and more configurable.

### DIFF
--- a/maple-core/src/easing.rs
+++ b/maple-core/src/easing.rs
@@ -103,43 +103,54 @@ pub fn circ_out(t: f32) -> f32 {
 
 pub fn circ_inout(t: f32) -> f32 {
     if t < 0.5 {
-        circ_in(2.*t) / 2.
+        circ_in(2. * t) / 2.
     } else {
-        (1. - (-2.*t + 2.).powi(2)).sqrt() / 2.
+        (1. - (-2. * t + 2.).powi(2)).sqrt() / 2.
     }
 }
 
 // Exponential
 
 pub fn expo_in(t: f32) -> f32 {
-    if t.abs() <= 0.0002 {0.}
-    else {EXP_BASE.powf(10.*t - 10.)}
+    if t.abs() <= 0.0002 {
+        0.
+    } else {
+        EXP_BASE.powf(10. * t - 10.)
+    }
 }
 
 pub fn expo_out(t: f32) -> f32 {
-    if (t - 1.).abs() <= 0.0002 {0.}
-    else {1. - EXP_BASE.powf(-10.*t)}
+    if (t - 1.).abs() <= 0.0002 {
+        0.
+    } else {
+        1. - EXP_BASE.powf(-10. * t)
+    }
 }
 
 pub fn expo_inout(t: f32) -> f32 {
-    if t.abs() <= 0.0002 {0.}
-    else if (t - 1.).abs() <= 0.0002 {1.}
-    else if t <= 0.5 {EXP_BASE.powf(20.*t - 10.) / 2.}
-    else {2. - EXP_BASE.powf(-20.*t + 10.) / 2.}
+    if t.abs() <= 0.0002 {
+        0.
+    } else if (t - 1.).abs() <= 0.0002 {
+        1.
+    } else if t <= 0.5 {
+        EXP_BASE.powf(20. * t - 10.) / 2.
+    } else {
+        2. - EXP_BASE.powf(-20. * t + 10.) / 2.
+    }
 }
 
 // Sine
 
 pub fn sine_in(t: f32) -> f32 {
-    1. - (t*PI/2.).cos()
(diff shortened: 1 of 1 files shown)


lukechu10:
Hi there. Thanks for taking the time to contribute to Maple! About testing the easing functions, I plan to add something similar to https://svelte.dev/examples#easing which should allow to visually test the easing functions. If I can get that finished fast enough, you can rebase it.

lukechu10:
I thought about testing and a simple sanity test could just be checking that `ease(0.0) == 0.0` and `ease(1.0) == 1.0`.

A macro could generate these test cases for all the easing functions.

maple-core/src/easing.rs
```diff
@@ -84,4 +91,98 @@ pub fn quint_inout(t: f32) -> f32 {
...
+    1. - (1. - t.powi(2)).sqrt()
+}
+
+pub fn circ_out(t: f32) -> f32 {
+    (1. - (t - 1.).powi(2)).sqrt()
+}
+
+pub fn circ_inout(t: f32) -> f32 {
+    if t < 0.5 {
+        circ_in(2. * t) / 2.
```
lukechu10, line 106:
```suggestion
        (1. - (1. - (2. * t).powi(2)).sqrt()) / 2.
```

maple-core/src/easing.rs
```diff
@@ -84,4 +91,98 @@ pub fn quint_inout(t: f32) -> f32 {
...
+    } else {
+        (1. - (-2. * t + 2.).powi(2)).sqrt() / 2.
```
lukechu10, line 108:
```suggestion
        ((1. - (-2. * t + 2.).powi(2)).sqrt() + 1.) / 2.
```

maple-core/src/easing.rs
```diff
@@ -84,4 +91,98 @@ pub fn quint_inout(t: f32) -> f32 {
...
+
+pub fn expo_inout(t: f32) -> f32 {
+    if t.abs() <= 0.0002 {
+        0.
+    } else if (t - 1.).abs() <= 0.0002 {
+        1.
+    } else if t <= 0.5 {
+        EXP_BASE.powf(20. * t - 10.) / 2.
+    } else {
+        2. - EXP_BASE.powf(-20. * t + 10.) / 2.
```
lukechu10, line 138:
```suggestion
        1. + EXP_BASE.powf(-20. * t + 10.) / -2.
```

maple-core/src/easing.rs
```diff
@@ -84,4 +91,98 @@ pub fn quint_inout(t: f32) -> f32 {
...
+        circ_in(2. * t) / 2.
+    } else {
+        (1. - (-2. * t + 2.).powi(2)).sqrt() / 2.
+    }
+}
+
+// Exponential
+
+pub fn expo_in(t: f32) -> f32 {
+    if t.abs() <= 0.0002 {
```
lukechu10, line 115: There should be some kind of constant like `EPSILON` instead of hard coding the error margin everywhere.

lukechu10 (changes requested):
I got a few nits if you don't mind. I would prefer if all the numbers had at least 1 digit after the decimal point (so `1.0` instead of `1.`). I think this makes it easier to see that it's a float rather than a number with a function call. It might also be clearer if you used associated functions (so `f32::powi(t - 1.0, 2)` instead of `(t - 1.0).powi(2)`).

I tried out the easing functions in the `tweened` example (you can run it using `trunk serve`) and some of them didn't work as expected so I added some fixes.

maple-core/src/easing.rs
tmpr: Hey, I'm kinda new and I pressed "Resolve" on all of the above suggested changes; can I somehow have them on my machine locally? Because I don't think they are present, probably because the changes you made live on your branch.

Also, I'm totally happy to further change the code, i.e. change the float-"format" and change the functions.

lukechu10: You can go into the "Files Changed" tab at the top and click on "Add suggestion to batch" for each suggestion. You should then be able to commit the suggestions. Finally, run `git pull` on your local machine. Hope this helps.

tmpr: This helped a lot! Thank you!

tmpr:
I switched out the formats now, but again this is just formalities so I hope you like it this way. :)

lukechu10 (approved):
Looks good. I'll merge this now

PR #158, opened 2021-07-08
Fix: use push_str instead of format

### LINKED ISSUE #154: Remove `format!` calls in `SsrNode`
Using `format!` is bad for performance because it pulls in the formatting infrastructure. Instead, replace all instances where `format!` is used with `push` and `push_str`. Also use `String::with_capacity` when possible to avoid unnecessary allocations.

rbrishabh:
Fixes #154
Hi @lukechu10! this is my first time working with rust. Could you tell me if this is the correct way to do it?
if yes, I can do the other ones too! thanks

### DIFF
--- a/packages/sycamore-macro/src/template/attributes.rs
+++ b/packages/sycamore-macro/src/template/attributes.rs
@@ -135,7 +135,7 @@ impl ToTokens for Attribute {
                         tokens.extend(
                             syn::Error::new(
                                 prop.span(),
-                                &format!("property `{}` is not supported with bind:", prop),
+                                "property `{}` is not supported with bind:".push_str(prop),
                             )
                             .to_compile_error(),
                         );


lukechu10:
Hello there. Thanks for taking the time to contribute to Sycamore!

As you can see, the code doesn't compile because you are calling `push_str` on `&str` instead of `String`. Since this is your first time working with Rust, I would strongly recommend that you first become familiar with the concepts described here: https://doc.rust-lang.org/book/ch04-03-slices.html

In a nutshell, `String` is an owned type whereas `&str` is a slice (if you are familiar with C++, basically a `std::string_view`, if not, it's a pointer/reference + length pair).

You can only call `push_str` on a `String` because it changes the length of the string.

What I want to do is something like:
```rust
let adjective = ADJECTIVES[random(ADJECTIVES_LEN_F64)];
let colour = COLOURS[random(COLOURS_LEN_F64)];
let noun = NOUNS[random(NOUNS_LEN_F64)];
let capacity = adjective.len() + colour.len() + noun.len() + 2;
let mut label = String::with_capacity(capacity);
label.push_str(adjective);
label.push(' ');
label.push_str(colour);
label.push(' ');
label.push_str(noun);
```
This prevents unnecessary allocations because the capacity of the `String` is known before hand.

And lastly, don't remove usage of `format!` inside `packages/sycamore-macro`. These are called rarely, and only during compile-time and so do not need to be optimized. I would prioritize readability over performance here. Only files in `packages/sycamore/generic_node/ssr_node.rs` could use this optimization because they are part of the hot path.

If you have questions or need some help, don't hesitate to ask me.

rbrishabh:
makes sense! working on this over the weekend :)
Thank you!

rbrishabh:
hi! I have a busy schedule and cant keep up with this now! Please feel free to continue it from here. I can close it for now

PR #162, opened 2021-07-10
fix intra-doc link

tshepang:


packages/sycamore/src/rx.rs
```diff
@@ -11,7 +11,7 @@ pub use motion::*;
 pub use signal::*;
 
 /// Creates a new reactive root / scope. Generally, you won't need this method as it is called
-/// automatically in [`render`](crate::utils::render()).
+/// automatically in [`utils::render`][crate::utils::render].
```
lukechu10, line 14: Actually, I meant
```suggestion
/// automatically in [`render`](crate::render).
```

lukechu10:
I'm not sure if you meant to apply my suggestions when you force-pushed but they are not there.

tshepang:
I should have, just forgot I could just accept them. Fixed now.

Issue #129, opened 2021-06-27
Router should not prevent default if meta keys are held down (e.g. `Ctrl + Click`)


lukechu10:

baile320:
I can help with this one, it seems like there just needs to be a couple ```if``` statements in ```router.rs``` where ```ev.prevent_default()``` is called, and we'll check whether ```let ctrl_key_pressed: bool = ev.clone().unchecked_into::<KeyboardEvent>().meta_key();``` is true/false?

lukechu10:
> I can help with this one, it seems like there just needs to be a couple `if` statements in `router.rs` where `ev.prevent_default()` is called, and we'll check whether `let meta_key_pressed: bool = ev.clone().unchecked_into::<KeyboardEvent>().meta_key();` is true/false?

Go ahead! There are also a few other meta keys which need to be checked for as well: meta, ctrl, shift and alt. The router in SvelteKit might be a good reference:
https://github.com/sveltejs/kit/blob/8854e2ffc48cf653854828f6ac19b2ce620535a6/packages/kit/src/runtime/client/router.js#L103

PR #165, opened 2021-07-10
Router should not prevent default if meta keys are held down

baile320:
This PR should close https://github.com/sycamore-rs/sycamore/issues/129.

I'm not 100% sure of the logic, because in the SvelteKit router code that was referenced in the issue (https://github.com/sveltejs/kit/blob/8854e2ffc48cf653854828f6ac19b2ce620535a6/packages/kit/src/runtime/client/router.js#L103), the key presses were checked immediately and if found the function immediately returned. So, in the code below, I skipped all of the subsequent logic if any of those keys were pressed. I'm not sure if the behavior is that we expect only the ```prevent_default``` from happening (in which case, I can update those two sections separately).

### DIFF
--- a/packages/sycamore-router/Cargo.toml
+++ b/packages/sycamore-router/Cargo.toml
@@ -24,9 +24,9 @@ features = [
   "EventTarget",
   "History",
   "HtmlAnchorElement",
+  "KeyboardEvent",
   "Location",
   "PopStateEvent",
   "Window",
-  "KeyboardEvent",
 ]
 version = "0.3"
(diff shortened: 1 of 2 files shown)


packages/sycamore-router/Cargo.toml
```diff
@@ -27,5 +27,6 @@ features = [
   "Location",
   "PopStateEvent",
   "Window",
+  "KeyboardEvent",
```
lukechu10, line 30: Nit: could you please specify the features in alphabetical order

lukechu10 (approved):
lgtm.
We don't want `prevent_default` from running because that would override the browser's built-in behavior when these keys are pressed.

packages/sycamore-router/Cargo.toml
baile320: done!

lukechu10:
Thanks for contributing!

PR #170, opened 2021-07-12
Explicitly enable std feature for indexmap

Gearme:
In some environments, indexmap's detection of the std crate fails, causing sycamore builds to fail (IndexSet has a variable number of type arguments depending on this detection).
This pull request sets indexmap's std feature flag explicitly, negating the need for automatic detection.

See also:
* [indexmap issue](https://github.com/bluss/indexmap/issues/184)
* [similar issue in google/cargo-raze](https://github.com/google/cargo-raze/issues/114)
* [indexmap's has_std causing issues with intellij](https://github.com/intellij-rust/intellij-rust/issues/4631)


lukechu10:
Thanks for contributing!

PR #180, opened 2021-07-20
fix logic error in `reconcile_fragments`

dicej:
I noticed when using the `Keyed` component that when the `KeyedProps::iterable`
value changed from a large `Vec` to a smaller one containing a subset of the
original, not all the obsolete nodes were removed from the DOM.

Specifically, I started with a `Vec` with 10 elements, then removed all but the
third one.  The nodes before the remaining node were removed, but not the ones
after it.  This patch fixes the problem.


lukechu10:
Thanks for catching this!

PR #195, opened 2021-08-01
set MSRV (to 1.53)

tshepang:


tshepang:
I included testing on macOS and Windows, for I was curious

README.md
```diff
@@ -18,6 +18,11 @@ Learn more at the [Sycamore website](https://sycamore-rs.netlify.app), or stop b
 Issue reports and Pull Requests are always welcome!
 Check out the [section on contributing](https://sycamore-rs.netlify.app/contribute/architecture) in the docs.
 
+## MSRV
+
+The minimum supported Rust toolchain is v1.53.0,
+due to usage of or-patterns syntax.
```
lukechu10, line 24:
```suggestion
The minimum supported Rust toolchain is v1.53.0. Sycamore is not guaranteed to compile on an older version of Rust.
```

.github/workflows/test.yml
```diff
@@ -7,8 +7,18 @@ env:
...
-    name: Test
-    runs-on: ubuntu-latest
+    name: Test on ${{ matrix.os }} (using rustc ${{ matrix.rust }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        rust:
+          - 1.53.0
+          - stable
+          - nightly
```
lukechu10, line 17: GitHub won't let me comment on an unchanged line but I think you missed replacing `toolchain: stable` with `toolchain: ${{matrix.rust }}` on line 29.

Also the cache key on line 47 should be updated with the rust version because different versions of rust are incompatible.
Something like:
```yaml
key: ${{ runner.os }}-cargo-${{ matrix.rust }}-${{ hashFiles('**/Cargo.lock') }}
```

README.md
tshepang: Bumping MSRV is an explicit (and somewhat rare) thing, and one that does it would know why, and would not miss updating this. An alternative is to only care about supporting latest `stable`, and that's when I think the reason is not needed.

lukechu10: Sorry, I wasn't being clear here. What I meant was that or-patterns might not be the only 1.53 feature we use. I seem to recall that `RefCell::take` and a few other functions were also stabilized recently.

lukechu10:
What I meant about `trybuild` in the issue was that rustc error messages tend to vary a lot between releases. Therefore the ui tests should only be run on 1.53.0

A simple fix for this would be to add the following to the tests in `packages/sycamore-macro/test/template_macro.rs` and `packages/sycamore-router-macro/router-macros.rs`:
```rust
#[test]
fn ui() {
    if std::env::var("RUN_UI_TESTS").is_ok() {
        // stuff here
    }
}
```
And only set the environment variable when `matrix.os == 1.53.0`

README.md
tshepang: The only error I get when compiling with 1.52 is about or-patterns.

lukechu10: But setting our MSRV to 1.53.0 does not mean that or-patterns is the only feature we are allowed to use. And technically, we could remove or-patterns from the code base to make it compatible with older versions of Rust. It's not like it's as crucial as some other features like `specialization` or `fn_traits`.

lukechu10:
Awesome. Thanks!

PR #204, opened 2021-08-07
Extract reactive primitives into new crate

lukechu10:


djmitche:
Will this crate have a fairly stable API?  For example, if I wanted to build a curses frontend with this new crate; or to build a UI for a microcontroller-based device; or maybe build an IoT hub on top of this?

lukechu10:
At the moment, non of the crates have "stable APIs" since we're still at the `0.x` stage. However, all the major version numbers are synchronized between the crates so this crate is just as stable as the `sycamore` crate if you will.

PR #206, opened 2021-08-09
Explain that Trunk needs a HTML source file

Yoshanuikabundi:
Hi! Just exploring Sycamore (looks cool so far) and ran into trouble running the hello world example, which doesn't describe creating index.html, which Trunk needs. Following the current instructions gives an error:

```
$ trunk serve                        
Error: error getting canonical path to source HTML file "index.html"

Caused by:
    No such file or directory (os error 2)
```

So I thought I'd fix that :)

The page in question also previously said:

> In fact, the Hello World program in Sycamore is but slightly longer than the console version!

But the console version is not described. So I've removed this sentence.

Finally, the description of the sycamore repo on Github is

> A reactive DOM library for Rust in WASM 

I only mention this in case you wanted to update it to "Sycamore is a modern VDOM-less web library with fine-grained reactivity in Rust and WebAssembly." to match the readme.

Thanks!

### DIFF
--- a/docs/versioned_docs/v0.5/getting_started/hello_world.md
+++ b/docs/versioned_docs/v0.5/getting_started/hello_world.md
@@ -1,7 +1,6 @@
 # Hello, World!
 
-Sycamore tries to have as simple of an API as possible. In fact, the Hello World program in Sycamore
-is but slightly longer than the console version!
+Sycamore tries to have as simple of an API as possible.
 
 Here it is:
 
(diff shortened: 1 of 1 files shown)


lukechu10:
LGTM. There definitely needs some consistency between the Sycamore taglines. Maybe "blazing-fast Rust library for creating web applications (with WebAssembly)" or something of the like.

PR #248, opened 2021-09-17
Update docs to use boolean for disabled attribute on button

redian:
Going through [docs](https://sycamore-rs.netlify.app/docs/basics/template#attributes) came across this 
and Rust compiler complained, hence this update.

Screenshot 2021-09-17 at 16 36 24


lukechu10:
Thanks for the contribution! BTW, I reverted the change on v0.5 docs because this is specific to v0.6

PR #253, opened 2021-09-19
Added docs on topological dependency tracking

### LINKED ISSUE #252: Documentation about how dependency tracking works
Dependency tracking is topological, meaning that dependencies must be accessed before the tracking scope returns.

```rust
create_effect(move || {
    // blah blah...
    
    wasm_bindgen_futures::spawn_local(async move {
        // This scope is not tracked because spawn_local runs on the next microtask tick (in other words, some time later).
    };
    // Everything that is accessed until here is tracked. Once this closure returns, nothing is tracked.
});
```

This means that accessing signals or other dependencies inside something like `wasm_bindgen_futures::spawn_local` will not track the dependencies and might not be what the author intended.

This should be clarified in the documentation about "Reactivity"

arctic-hen7:
Closes #252.

### DIFF
--- a/docs/versioned_docs/v0.6/basics/reactivity.md
+++ b/docs/versioned_docs/v0.6/basics/reactivity.md
@@ -147,7 +147,7 @@ We'll find that any `Signal`s we track in the `create_effect` won't be tracked p
 
 ```rust
 create_effect(move || {
-	let _ = signal.get(); // Where `signal` is a reactive dependency
+    let _ = signal.get(); // Where `signal` is a reactive dependency
     wasm_bindgen_futures::spawn_local(async move {
         // This scope is not tracked because spawn_local runs on the next microtask tick (in other words, some time later).
     };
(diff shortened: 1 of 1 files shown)


docs/versioned_docs/v0.6/basics/reactivity.md
```diff
@@ -127,3 +127,32 @@ let root = {
...
+    };
+    // Everything that is accessed until here is tracked. Once this closure returns, nothing is tracked.
+});
+```
+
+We'll find that any `Signal`s we track in the `create_effect` won't be tracked properly in the `wasm_bindgen_futures::spawn_local`, which is often not what's intended. This problem can be gotten around by accessing reactive dependencies as needed before going into a future, or with this simple fix:
+
+```rust
+create_effect(move || {
+	let _ = signal.get(); // Where `signal` is a reactive dependency
```
lukechu10, line 150: Seems like this is a tab rather than 4 spaces.
```suggestion
    let _ = signal.get(); // Where `signal` is a reactive dependency
```

lukechu10:
Thanks a lot!

PR #257, opened 2021-09-21
impl Default for Signal<T: Default>

### LINKED ISSUE #259: Consider implementing Default for Signal<T: Default>
It would be very convenient, especially for component props which take Signal's to implement Default, to avoid having to manually implement Default for all prop fields.

#xample:
```rust
#[derive(Default)]
pub struct ComponentProps {
    reactive_prop: Signal<Option<()>>
}
```

instead of
```rust
impl Default for ComponentProps {
    /* ... */
}
```

jquesada2016:
Closes #259

### DIFF
--- a/packages/sycamore-reactive/src/signal.rs
+++ b/packages/sycamore-reactive/src/signal.rs
@@ -198,6 +198,15 @@ impl<T: 'static> Signal<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


lukechu10:
LGTM, Thanks!

Issue #264, opened 2021-09-23
Do not implicitly call `.to_string()` in `view!`


jquesada2016:
I was trying to use a bool attribute for an `<a>` tag as follows:

```rust
let link_text = "blah";
let download = false;

template! {
    a(download=download) { (link_text) }
}
```

And when I inspected the `DOM`, I noticed the `download` attribute was set to `"false"`. Looking at the `cargo expand` output, I noticed that all, but a few attributes, such as `required` were being treated specially.

I believe it might be a good idea to implement a trait such as the following (pseudo code):
```rust
trait AttributeKind {
    const fn is_bool() -> bool
}

impl AttributeKind for bool {
    const fn is_bool() -> bool { true }   // false for string
}
```

Which could then be used in a macro as follows:
```rust
/* inside of template! macro */
attribute!(value)
```

Which could expand to _something_ like:
```rust
if (value as AttributeKind)::is_bool() {
    /* add the required code to add/remove the attribute
} else {
    /* use the normal ToString trait */
{
```

This could work for all possible attributes and wouldn't have to be cherry-picked.

The trait could be only really have to be implemented for bool, returning `true`, and a blanket implementation for anything implementing ToString returning `false`.

jquesada2016:
On this note, accepting Option would also be a good idea, because when working with Web Components, many of their API's involve adding/removing attributes to toggle or set functionality. I don't currently see a way to toggle attributes without using the raw Template and GenericNode API's.

lukechu10:
According to the [spec](https://html.spec.whatwg.org/#attributes-3), download is a text attribute.
To set `download` without specifying a value, you can use the empty string slice (`""`).

I'm beginning to think that maybe it's not such a good idea to use the `ToString` trait for all types as it can result in somewhat unexpected behavior at times. It might be better to only allow `AsRef<str>` values.

jquesada2016:
I recommend checking out [Seed](https://github.com/seed-rs/seed) and how they go about this in their `HTML` macros. That's where I got the idea for adding an `IntoAttributes` trait. Ideally, one would have an `IntoAttribute` and an `IntoChild` (or whatever) for each respective part of the `template!`. Then one could implement it for the important types (`bool`, `&str`, etc.), and this would give complete predictability and better error reporting. The main blocker right now is the stabilization of `specialization`. Or we could use the `Any` trait if using `'static`...

lukechu10:
> Or we could use the `Any` trait if using `'static`...

[...]

PR #265, opened 2021-09-24
Added node! convenience macro

jquesada2016:
This macro works well for cases where we need access to underlying `GenericNode` operations, such as `.event()`, `.add_attribute()`, `.remove_attribute` that cannot currently be performed inside of the `template!` macro, yet retain the ergonomics for setting attributes and children that don't need the extra flexibility. Please see the `doc example` below for a possible use case.

This is also a good workaround for #264.

### DIFF
--- a/packages/sycamore-macro/src/lib.rs
+++ b/packages/sycamore-macro/src/lib.rs
@@ -16,15 +16,13 @@ pub fn template(component: TokenStream) -> TokenStream {
 }
 
 /// ```
-/// use sycamore::{generic_node::EventHandler, prelude::*};
+/// use sycamore::prelude::*;
 ///
 /// #[component(MyComponent<G>)]
-/// pub fn my_component(event_listeners: Vec<(String, Box<EventHandler>)>) -> Template<G> {
+/// pub fn my_component() -> Template<G> {
 ///     let cool_button: G = node! { button { "The coolest 😎" } };
 ///
-///     for listener in event_listeners {
-///         cool_button.event(listener.0.as_ref(), listener.1);
-///     }
+///     cool_button.set_property("myProperty", &"Epic!".into());
 ///
 ///     Template::new_node(cool_button)
 /// }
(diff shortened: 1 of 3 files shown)


packages/sycamore-macro/src/lib.rs
```diff
@@ -15,6 +15,27 @@ pub fn template(component: TokenStream) -> TokenStream {
     template::template_impl(component).into()
 }
 
+/// ```
+/// use sycamore::{generic_node::EventHandler, prelude::*};
```
lukechu10, line 19:
```suggestion
/// use sycamore::generic_node::EventHandler;
/// use sycamore::prelude::*;
```

packages/sycamore-macro/src/lib.rs
```diff
@@ -15,6 +15,27 @@ pub fn template(component: TokenStream) -> TokenStream {
...
+///     Template::new_node(cool_button)
+/// }
+/// ```
+#[proc_macro]
+pub fn node(input: TokenStream) -> TokenStream {
+    let node = parse_macro_input!(input as template::Element);
+
+    template::node_impl(node).into()
+}
+
```
lukechu10, line 38: How about we do something else such as setting a property because this is technically bad practice: the listeners will never be dropped which is a memory leak.
```suggestion
/// A macro for creating nodes.
/// ```
/// pub fn my_component() -> Template<G> {
///     let cool_button: G = node! { button { "The coolest 😎" } };
///     cool_button.set_property("my_property", &123.into());
///     Template::new_node(cool_button)
/// }
/// ```
#[proc_macro]
pub fn node(input: TokenStream) -> TokenStream {
    let node = parse_macro_input!(input as template::Element);

    template::node_impl(node).into()
}
```

lukechu10 (commented):
lgtm. Just two nitpicks.

jquesada2016:
Applied suggested changes!

lukechu10:
The diff has some commits that are not supposed to be there. If you don't mind, I can fix this and I'll get this merged

jquesada2016:
I removed all commits that weren't supposed to be there. It was the 0.6.1 release that I pulled in. If there's still something that isn't supposed to be there, go right ahead!

lukechu10:
Sorry for the run-around but it seems like my suggested changes got undone.

jquesada2016:
No worries! Let me know what you would like me to do in this case.

PR #267, opened 2021-09-26
Added `remove_property` method to `GenericNode`

jquesada2016:
This is useful to have, especially for some serializers which require certain properties to not exist, not just be `undefined`.


lukechu10:
Thanks!

Issue #268, opened 2021-09-26
Builder Pattern API


jquesada2016:
### Proposal

I have been contemplating what a builder pattern API, as mentioned in #265, might look like. I personally like the idea of having both API's, and having the interop, as macros have their place which can be much easier to read, but the primary advantages, in my opinion to a builder pattern is that, currently, formatting and type hints don't work with the current tooling (rust-analyzer), and certain more complicated logic is not possible at the moment, without further syntax extensions.

Also, from a discord comment, @lukechu10 mentioned the aim of eventually supporting React Native, and possibly other rendering targets. This to me, presents two different routes that could be taken:

1. Having a rendering-agnostic API
2. Having a rendering-specific API

The first option could be something as follows (for HTML):

```rust
button();
div();
a();
/* etc */

```

Whereas the agnostic approach could be something like:

```rust
node("button");
node("div");
node("a");

```

These features could be held behind a feature flag, such as `html`, `react-native`, `native-script`, etc. for the render-specific API, or just `builder-api` for the agnostic API, and prefixed with `experimental:` until an API is stabilized.

### Main required features

In no particular order. Also note that all of the mentioned functions would be implemented for the builder struct and take either `&mut self` or `self` and return their respective `Self` type, but have been omitted for brevity..

- (Un)setting Attributes
    - A further distinction could be made (for memory and performance reasons) between static and dynamic attributes, such as:

        ```rust
        fn set_attribute(name: &str, value: impl Into<JsValue>);

        // The Option type is important here, as it can be used to automatically remove
        // the attribute when required.
        fn set_dyn_attribute(name: &str, value: StateHandle<Option<impl Into<JsValue>>>);

        // Also note that remove_attribute is not needed here, since it's handled in the
        // dynamic case and in static cases not needed.
        ```

- (Un)setting properties
    - Same as above
- (Un)setting classes
    - Once again, I recommend having a distinction between static and dynamic classes.

        ```rust
        fn add_class(class: &str);
        fn add_dyn_class(class: StateHandle<Option<&str>>);
        ```

- Add/remove children
    - For setting children, we have some choices to make. We can accept `Vec<GenericNode>`, or `Template`, thought the latter is preferable. The question lies on how to handle dynamic children.

        ```rust
        fn add_children(children: Template<G>);

        // Perhaps dynamic children could be handled as follows
        fn add_dyn_children(children: StateHandle<Template<G>>);
        ```

- Special shortcuts
    - `fn id(id: &str)`
[...]

PR #269, opened 2021-09-28
Builder API

jquesada2016:
I have not fully tested, nor do I think the current iteration has the right aesthetics, but I believe it's in the right direction! This is an implementation attempt for the builder API as mentioned in #268.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,3 +129,5 @@ jobs:
         run: |
           cd packages/sycamore
           cargo miri test
+          cd ../sycamore-reactive
+          cargo miri test
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -4,7 +4,7 @@ FROM gitpod/workspace-full
 RUN curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 
 # Install trunk
-RUN bash -cl "wget -qO- https://github.com/thedodd/trunk/releases/download/v0.13.1/trunk-x86_64-unknown-linux-gnu.tar.gz | tar -xzf-"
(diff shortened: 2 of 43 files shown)


jquesada2016:
@lukechu10 All tests should pass. I went through the logs and adjusted some code to run on rustc v1.53, but it seems the tests did not re-run. Not sure why. Also, API is stable enough for an initial audit. I've migrated our project to use the API as it currently stands and have no issues. Parts that remain untested are the `styles` functions, since we aren't using styles. I will also be adding docs to the functions.

lukechu10:
Wow thanks for all the work. I'll try to get to this soon.

jquesada2016:
Would you like to try to aim this feature for `v0.7.0`?

lukechu10:
Yes. That sounds good.

lukechu10:
> No problem! I'm going to be mostly MIA the next couple weeks due to traveling. But I'll try to work on this whenever I get a chance.

No worries!



> What I was thinking was eventually making this a public macro which users could use for extending the builder API to use custom typed tags, which would be very useful for quickly adding support for something like Svelte Native, along with web components UI frameworks. Also, since this is a very simple macro, it didn't merit, in my opinion, dropping to a full procedural macro.

What I meant was
```rust
generate_tag_functions! { enum HtmlTag { ... } }
```
instead of
```rust
#[apply(generate_tag_functions)]
enum HtmlTag { ... }
```
because they are equivalent.



> Sure, we can remove these, though I can imagine some workflows, especially with dynamic tables or virtual lists where this kind of optimization could be useful. Perhaps mark the function as unsafe? (though this might break the builder flow) If not, I'm fine with removing it.

We might be able to do something about this with type-states but I think that belongs in another PR.

lukechu10:
If you don't mind, I can merge this now and fix the issues I mentioned in a later PR to not make this one too long.

jquesada2016:
Go right ahead!

PR #274, opened 2021-10-09
Add `IS_BROWSER` constant to `GenericNode`

lukechu10:
This fixes the ugly hack of using
```rust
if TypeId::of::<G>() == TypeId::of::<DomNode>() { ... }
```
and replaces it with
```rust
if G::IS_BROWSER { ... }
```

### DIFF
--- a/packages/sycamore-macro/src/component/mod.rs
+++ b/packages/sycamore-macro/src/component/mod.rs
@@ -168,7 +168,7 @@ pub fn component_impl(
     let component_name_str = component_name.to_string();
     let generic_node_ty = generic_node_ty.type_params().next().unwrap();
     let generic_node: GenericParam = syn::parse_quote! {
-        #generic_node_ty: ::sycamore::generic_node::GenericNode
+        #generic_node_ty: ::sycamore::generic_node::Html
     };
 
     let ComponentFunction {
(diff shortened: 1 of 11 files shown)


jquesada2016:
@lukechu10 Since you mentioned wanting to support other platforms in the future, wouldn't it be wiser to change the IS_BROWSER constant to something like PLATFORM_NAME or PLATFORM with an enum which can include a variant for custom for user provided renderers? This does solve the ugly hack, but doesn't help when targeting other runtimes, such as NativeScript, React Native, etc.

lukechu10:
> @lukechu10 Since you mentioned wanting to support other platforms in the future, wouldn't it be wiser to change the IS_BROWSER constant to something like PLATFORM_NAME or PLATFORM with an enum which can include a variant for custom for user provided renderers? This does solve the ugly hack, but doesn't help when targeting other runtimes, such as NativeScript, React Native, etc.

Only web apps need to be isomorphic, meaning that they should render with both `DomNode` and `SsrNode`. For other platforms, there will only ever be one single backend and so this will not be used. In fact, now that I think about it, maybe the constant should go on the new `Html` trait (that is only implemented for `DomNode` and `SsrNode`) instead of `GenericNode`.

PR #278, opened 2021-10-14
Prevent binding from panicking when not in wasm

zyllian:
This fixes an issue I ran across while using the same template code to do SSR and DOM rendering where using `value:bind` would panic during SSR because it tries to call an unimplemented function from `wasm_bindgen`.

This isn't a particularly elegant solution to the problem, so I'm sure it can be improved.


arctic-hen7:
Indeed, this will make using data binding possible in Perseus!

lukechu10:
Ah, the tests are failing because the `#[cfg]`s are in the wrong place. These are evaluated during compile time so the target is the target on which the code is being built. The `#[cfg]`s should belong inside the `quote!` macro instead.

PR #284, opened 2021-10-16
Add helper macros for common functions that use `cloned!`

emi2k01:
This commit adds a macro for generating wrappers around functions that
accept a closure and makes it easier to pass cloned variables.

This commit also uses said macro to generate wrappers for the following
functions:

- `create_effect`
- `create_memo`
- `create_selector`
- `create_root`

### DIFF
--- a/docs/next/advanced/routing.md
+++ b/docs/next/advanced/routing.md
@@ -277,7 +277,7 @@ use sycamore::futures::spawn_local_in_scope;
 template! {
     Router(RouterProps::new(HistoryIntegration::new(), |route: StateHandle<AppRoutes>| {
         let template = Signal::new(Template::empty());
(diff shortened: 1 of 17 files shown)


lukechu10:
Hmm, it seems like we would need to update MSRV to 1.54 to make this possible (which is fine by me).

lukechu10:
I've given a few more thoughts about this and I don't think this is a good idea anymore. Some hooks also take arguments and that would be difficult with the macro syntax. I am thinking of create an alias for the `cloned!` macro, probably just `c!` so that it's a bit shorter.

```rust
create_effect!(my_signal, move || {
    my_signal.get();
});
```
vs.
```rust
create_effect(c!(my_signal, move || {
    my_signal.get();
}));
```

Issue #279, opened 2021-10-15
Update TodoMVC example to use Context API


lukechu10:
The TodoMVC example is a good reference for learning Sycamore. It should be updated to use the [context API](https://sycamore-rs.netlify.app/docs/v0.6/advanced/contexts) to prevent "prop-drilling"

lukechu10:
awesome!

lukechu10:
No worries

PR #295, opened 2021-11-06
Update TodoMVC example to use Context API

haywoood:
Does what it says on the tin, and it's so beautiful 😭

Closes https://github.com/sycamore-rs/sycamore/issues/279


lukechu10:
Wow that was fast. Thanks a lot. Busy right now but will review soon

lukechu10 (approved):
Looks awesome!

PR #303, opened 2021-11-22
render_temporarily_to() helper

alsuren:
As I mentioned in https://discord.com/channels/820400041332179004/821489793258487839/912364912624418846 I was having problems with things not getting dropped when gradually converting an existing app to use sycamore.

This function is intended to be used for injecting an ephemeral sycamore view into a non-sycamore app
(for example, a file upload modal where you want to cancel the upload if the modal is closed).
You should only use this function while refactoring your app, and you should aim to have a single
call to [`render`] or [`render_to`] at the top level of your app long-term.

Using this function did solve my problems.

### DIFF
--- a/packages/sycamore/src/generic_node/dom_node.rs
+++ b/packages/sycamore/src/generic_node/dom_node.rs
@@ -289,6 +289,20 @@ pub fn render(template: impl FnOnce() -> View<DomNode>) {
     render_to(template, &document.body().unwrap_throw());
 }
 
+/// Render a [`View`] under a `parent` node.
+/// For rendering under the `<body>` tag, use [`render`] instead.
+///
+/// _This API requires the following crate features to be activated: `dom`_
+pub fn render_to(template: impl FnOnce() -> View<DomNode>, parent: &Node) {
+    let scope = render_get_scope(template, parent);
+
+    thread_local! {
+        static GLOBAL_SCOPES: std::cell::RefCell<Vec<ReactiveScope>> = std::cell::RefCell::new(Vec::new());
+    }
+
+    GLOBAL_SCOPES.with(|global_scopes| global_scopes.borrow_mut().push(scope));
+}
+
 /// Render a [`View`] under a `parent` node, in a way that can be cleaned up.
 /// This function is intended to be used for injecting an ephemeral sycamore view into a non-sycamore app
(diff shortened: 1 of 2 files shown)


packages/sycamore/src/generic_node/dom_node.rs
```diff
@@ -289,6 +289,34 @@ pub fn render(template: impl FnOnce() -> View<DomNode>) {
...
+/// This function is intended to be used for injecting an ephemeral sycamore view into a non-sycamore app
+/// (for example, a file upload modal where you want to cancel the upload if the modal is closed).
+/// You should only use this function while refactoring your app, and you should aim to have a single
+/// call to [`render`] or [`render_to`] at the top level of your app long-term.
+/// For rendering a view that will never be unmounted from the dom, use [`render_to`] instead.
+/// For rendering under the `<body>` tag, use [`render`] instead.
+///
+/// _This API requires the following crate features to be activated: `dom`_
+#[must_use = "please hold onto the ReactiveScope until you want to clean things up, or use render_to() instead"]
+pub fn render_temporarily_to(
```
lukechu10, line 302: I'm not sure this is the right name to use here. How about something like `render_get_scope`?

packages/sycamore/src/generic_node/dom_node.rs
```diff
@@ -289,6 +289,34 @@ pub fn render(template: impl FnOnce() -> View<DomNode>) {
...
+                id: Default::default(),
+                node: parent.clone(),
+            },
+            template(),
+            None,
+            None,
+            false,
+        );
+    })
+}
```
lukechu10, line 318: All this is duplicated in `render_to`. Maybe replace `render_to` with a call to this function.

lukechu10 (commented):
Thanks for the contribution. Just a few nitpicks.

packages/sycamore/src/generic_node/dom_node.rs
alsuren: I have pushed a change to use `render_get_scope`.

Other potential names include `render_and_get_scope` `render_and_return_scope` `render_to_scope` `mount_and_get_scope`. Shout if you prefer any of those.

lukechu10:
lgtm. thanks!

PR #305, opened 2021-11-28
An HTTP request example and a tailwindcss styling example

attriaayush:
I created this example (since I could not find an existing one) to get a feel for the developer experience of using a crate like [reqwasm](https://github.com/hamza1311/reqwasm) and styling with [TailwindCSS](https://tailwindcss.com/) when using sycamore.

This example makes a `GET` request to [CountAPI](https://countapi.xyz/) that simply increments everytime you visit the page or refresh the page i.e. counting page visits. I picked this example since I just wanted to create a really simple HTTP request to fetch some `json` data and then be able to display that data which then further updates a `Signal` and resulting into updating the relevant component. 

This also includes a little styling with TailwindCSS since the [docs here](https://sycamore-rs.netlify.app/docs/advanced/css) suggested that the styling part is still under works.

To test, you can run `trunk serve --open` which should build everything and the serve the page.


lukechu10:
I think it would be better to separate http requests using `reqwasm` and styling using `tailwindcss` into 2 distinct examples instead of combining them into one. What do you think?

attriaayush:
> I think it would be better to separate http requests using `reqwasm` and styling using `tailwindcss` into 2 distinct examples instead of combining them into one. What do you think?

Thanks for reviewing the pull request.

That makes sense and I agree. More than happy to split this example into two distinct examples.

attriaayush:
@lukechu10 
I have split this into two distinct examples. Still using the same base of http request example using `reqwasm` but this time isolating  the `tailwindcss` styling setup into a different example that just simply builds on top of the previous example with a unique `site_id`.

Hope these are useful examples.

examples/page-visit-counter-tailwindcss/Trunk.toml
```diff
@@ -0,0 +1,9 @@
+[build]
+target = "index.html"
+dist = "dist"
+
+[[hooks]]
+stage = "build"
+command = "sh"
+command_arguments = ["-c", "npx tailwindcss -o ./dist/.stage/tailwind.css"]
```
lukechu10, line 8:
```suggestion
command_arguments = ["-c", "npx -y tailwindcss -o ./dist/.stage/tailwind.css"]
```

examples/page-visit-counter-tailwindcss/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link href="/tailwind.css" type="text/css" rel="stylesheet" />
+    <title>Http Example App</title>
+
+    <link data-trunk rel="scss" href="index.scss" />
```
lukechu10, line 10:
```suggestion
    <link data-trunk rel="scss" href="index.scss" />
```
We can probably delete this since we're using `tailwind.css` already.

examples/page-visit-counter-tailwindcss/src/main.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[component(App<G>)]
+fn app() -> View<G> {
+    let counter = Signal::<u64>::new(0);
+
+    create_effect(cloned!((counter) => move || {
+        spawn_local_in_scope(cloned!((counter) => async move {
+            let website_id = "page-visit-counter-tailwindcss.tyz";
+            let visits = fetch_visits(website_id).await.unwrap();
```
lukechu10, line 49: Maybe instead of `.unwrap()`ing, use `.unwrap_or_default()` and derive `Default` for `Visits` so that the app doesn't crash if the API is not accessible.

examples/page-visit-counter/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+#[component(App<G>)]
+fn app() -> View<G> {
+    let counter = Signal::<u64>::new(0);
+
+    create_effect(cloned!((counter) => move || {
+        spawn_local_in_scope(cloned!((counter) => async move {
+            let website_id = "page-visit-counter.tyz";
+            let visits = fetch_visits(website_id).await.unwrap();
```
lukechu10, line 45: See previous comment

lukechu10 (commented):
I'm not sure if it's best to have the tailwindcss example also include http request but I think it's fine for now.

lukechu10:
Also CI is failing because latest reqwasm uses edition 2021 which needs rust 1.56 and sycamore's MSRV is currently 1.53.
I sent a PR to reqwasm changing edition back to 2018.

attriaayush:
@lukechu10 
Appreciate the feedback, I believe that all should be addressed now.

> I'm not sure if it's best to have the tailwindcss example also include http request but I think it's fine for now.

My thinking behind this was that tailwindcss is just a styling layer on top of an example and it could potentially be any existing example that we can build upon. In this case now that we have an isolated example for http why not just extend it and style the relevant component. Also majority of the time projects do contain http requests and styling thought it might represent that side of it in a really simple form.

lukechu10 (approved):
lgtm. I'm just waiting for reqwasm to push a new release so that it's compatible with rust 1.53

lukechu10:
I removed the non-tailwind example because in retrospect, I think it's better to not have duplicated code

PR #312, opened 2021-12-07
Filter phantom_generics to contain types only

### LINKED ISSUE #311: Const generic component compile error
Components with const generics fail to compile on master, Rust 1.57, 2021 edition.
```rust
#[component(MyComponent<G>)]
fn my_component<const N: usize>(i: [u8; N]) -> View<G> {
    View::new_fragment(
        i.into_iter()
            .map(|u| {
                view! {
                    (u)
                }
            })
            .collect(),
    )
}
```

The compiler gives the following error:
```rust
error[E0573]: expected type, found const parameter `N`
 --> examples/components/src/main.rs:4:23
  |
3 | #[component(MyComponent<G>)]
  |                         - similarly named type parameter `G` defined here
4 | fn my_component<const N: usize>(i: [u8; N]) -> View<G> {
  |                       ^ help: a type parameter with a similar name exists: `G`
```

... which is caused by the fact that const generic parameters are passed to `PhantomData`,  as seen in the following expanded output:
```rust
struct MyComponent<G: ::sycamore::generic_node::Html, const N: usize> {
    #[doc(hidden)]
    _marker: ::std::marker::PhantomData<(G, N)>,
}
```

Generic arguments for `PhantomData` need to be filtered to contain types only in the proc-macro.

danielalvsaaker:
Fixes #311.

### DIFF
--- a/packages/sycamore-macro/src/component/mod.rs
+++ b/packages/sycamore-macro/src/component/mod.rs
@@ -217,11 +217,18 @@ pub fn component_impl(
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
 
     // Generics for the PhantomData.
-    let phantom_generics = ty_generics
-        .clone()
-        .into_token_stream()
-        .into_iter()
-        .collect::<Vec<_>>();
+    let phantom_generics = {
+        // Filter out generics which are not types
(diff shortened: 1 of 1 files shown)


lukechu10 (approved):
Thanks!

PR #323, opened 2021-12-17
Make output of `render_to_string` fixed

### LINKED ISSUE #320: Stable render_to_string
When I add some test contains `sycamore::generic_node::ssr_node::render_to_string`, these tests randomly fails.

I found this is caused by `AHashMap`, if there is a stable output then it will be very useful for testing.

This is not used in a production environment, only for testing.

oovm:
Fix #320

### DIFF
--- a/packages/sycamore/src/generic_node/ssr_node.rs
+++ b/packages/sycamore/src/generic_node/ssr_node.rs
@@ -1,11 +1,7 @@
 //! Rendering backend for Server Side Rendering, aka. SSR.
 
-#[cfg(not(feature = "stable"))]
-use ahash::AHashMap;
 use once_cell::sync::Lazy;
 use std::cell::RefCell;
-#[cfg(feature = "stable")]
-use std::collections::BTreeMap;
 use std::collections::HashSet;
 use std::hash::{Hash, Hasher};
 use std::iter::FromIterator;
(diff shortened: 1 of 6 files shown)


packages/sycamore/src/generic_node/ssr_node.rs
```diff
@@ -1,19 +1,18 @@
...
-
-use ahash::AHashMap;
-use once_cell::sync::Lazy;
 use wasm_bindgen::prelude::*;
 
 use crate::generic_node::{GenericNode, Html};
 use crate::reactive::create_scope;
 use crate::utils::hydrate::{get_next_id, with_hydration_context};
 use crate::view::View;
+use indexmap::map::IndexMap;
```
lukechu10, line 15: Nit: Can you organize the imports here please? The order should be std, dependencies, self.

lukechu10 (commented):
lgtm

packages/sycamore/src/generic_node/ssr_node.rs
oovm: Sorry I didn't notice this, IDEA auto import breaks this part

lukechu10:
Seems like the build website workflow is broken. I guess I'll need to fix that.

Issue #302, opened 2021-11-17
Documentation rendering on mobile


ctjhoa:
**Describe the bug**
The book documentation have a rendering issue.

Background doesn't stretch over the whole width.

**To Reproduce**

1. Go to https://sycamore-rs.netlify.app/docs/getting_started/installation
2. Open dev tools and switch to mobile view (eg. iphone 11 size)
3. Scroll to the right

**Expected behavior**
Full width background

**Screenshots**
![Screenshot_20211117-192717.png](https://user-images.githubusercontent.com/1716173/142261449-e7988221-6cfc-4b14-ba5a-34325071bd60.png)

**Environment**

- Browser: Firefox 94.0

martinmcwhorter:
Is anyone working on this, if not I'll take a stab?

lukechu10:
> Is anyone working on this, if not I'll take a stab?

I think the issue here is really making the website more mobile friendly which would involve collapsing the nav menu into a hamburger menu. I got started a bit on the [`fix-website-css`](https://github.com/sycamore-rs/sycamore/tree/fix-website-css) branch but you're welcome to continue.

PR #331, opened 2022-01-12
fix broken website on small screens

xyning:
https://github.com/sycamore-rs/sycamore/issues/302  

sidebar & navbar are collapsed into the hamburger menu

side effects / fixes note:
* backdrop blurs now work in dark mode
* code blocks will scroll instead of wrapping if overflowed
* sidebar is pinned on desktop screen
* slightly increased blur radius for better text readability
* main text's max width is limited to around 900px
* slightly reduced the font size of sidebar
* added a small margin on index page

### DIFF
--- a/website/index.css
+++ b/website/index.css
@@ -12,7 +12,7 @@
   }
 
   .content pre {
-    @apply bg-gray-100 rounded block p-2 whitespace-pre-wrap font-mono;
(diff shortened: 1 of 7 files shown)


lukechu10:
Seems like I'll need to bump MSRV because of bumpalo. Thanks for the contribution!

PR #337, opened 2022-01-26
Reactive Primitives v2! 🎉

### LINKED ISSUE #107: Component children
Components (`#[component(_)]`) should be able to have child nodes or props.

## Design

Currently, component functions only accept an optional single parameter (the prop type). With this feature, component functions should allow for an optional second parameter that is the children passed to the component. To prevent ambiguity when the prop type is omitted, the children parameter must also be marked with a `#[slot]` attribute.

### Proposed changes to the `#[component(_)]` attribute macro
The proposed syntax is as follows:
```rust
#[component(MyComponent<G>)]
fn my_component(props: Props, #[slot] children: TemplateResult<G>) -> TemplateResult<G> {
    template! { ... }
}

// Or with props omitted (having type '()')
#[component(MyComponent<G>)]
fn my_component(#[slot] children: TemplateResult<G>) -> TemplateResult<G> {
    template! { ... }
}
```

### Children type and proposed changes to the `template!` macro
The proposed changes to the `template!` macro is as follows:
```rust
template! {
    MyComponent() {
        div
    }
}
```
This will pass a `TemplateResult` representing the `div` element to the component. Other types can be passed as well:
```rust
template! {
    MyComponent() {
        (123u32)
    }
}
```
This will pass a `u32` of value `123u32` to the component.

The macro should not attempt to do any casting of the type passed in.

## Additional information
This will improve the ergonomics of using `Indexed` and `Keyed` utility components. The map function can be passed as the children to reduce the number of props. The syntax would be as follows:
```rust
template! {
    Indexed(IndexedProps { each=my_vec }) {
        (|x| template! {
            p { (x) }
        })
    }
}
```

## Alternatives
Make `children` a field of the property type. This is the design of many other web frameworks such as React and Yew.
This does not fit as well into Rust patterns because Rust does not have structural types. One would therefore need an additional trait to access the child and make constructing components in the `template!` macro more complicated.

lukechu10:
Gets rid of all the clones!

Closes #107 

### Task list:
(some of these might get split up into separate PRs)
- [x] Port `sycamore-router`
- [x] Port website
- [x] Port SSR benchmark
- [x] Port `Portal`
- [x] Port tweening and motion
- [x] Update docs
- [x] Add more tests for `sycamore-reactive` because of all the `unsafe` code


arctic-hen7:
Congratulations and many thanks @lukechu10 for all your hard work on this!

PR #370, opened 2022-02-23
Builder doctests

alsuren:
There are a few tests that I have left as "compile_fail" because I don't fully understand how contexts and signals are expected to work in 0.8. Search for FIXME in the diff, to find them. Any guidance that can be provided on these points would be greatly appreciated.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,10 +52,10 @@ jobs:
         if: matrix.rust == '1.58.0'
         env:
           RUN_UI_TESTS: true
-        run: cargo test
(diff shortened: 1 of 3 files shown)


lukechu10:
Hello there. Sorry for taking so long to get to this.
With the new v2 reactive primitives, the old builder API didn't really play well with it, mainly because I kind of just changed the function signatures to make it compile without any attention to ergonomics.
I actually completely overhauled the builder API in #373 which makes the old builder API obsolete, so unfortunately, I'll be closing this PR now. Sorry again for all the work you put into this.
Thanks for understanding.

PR #376, opened 2022-03-06
Add some docs to `sycamore-futures` and deprecate `create_resource`

lukechu10:


t1m0t:
Add breaking_change label?

lukechu10:
I guess I could but doesn't really matter since this didn't exist in 0.7 and is new in 0.8

t1m0t:
> I guess I could but doesn't really matter since this didn't exist in 0.7 and is new in 0.8

Fair enough. Just though it would be useful but as you said there is no comparison to a previous release so nevermind 🙂

Issue #375, opened 2022-03-06
Add `dangerously_set_inner_html` support to builder API


lukechu10:
This should do the same thing as in the `view!` macro, e.g:

```rust
h(div)
    .dangerously_set_inner_html("<script>alert("unescaped html")</script>")
    .view(ctx)
```

The docs should also be updated in `docs/next/basics/view`

Issue #288, opened 2021-10-24
Add SVG Support


AMNRG:
Currently, SVG nodes are created without specifying a namespace URI and therefore do not work in the browser.
SVG nodes should be created with _createElementNS()_ instead of _createElement()_.

adundovi:
I've noticed that the only workaround is to use raw HTML:
https://github.com/sycamore-rs/sycamore/blob/fc640d313e66f9a6af422fae44f4f72fa86280cc/website/src/header.rs#L8
Right?

lukechu10:
Yeah, that's the workaround.

alsuren:
We were discussing this internally off the back of #378 (see my comment on that PR for other ways we've used for rendering SVGs)

One ideal-sounding api would be some kind of `fn svg_view(src: &str) -> View` method that's able to return an unwrapped svg. Typically you would want to call this as svg_view(include_bytes!("thing.svg")), because you want to be able to import/export/view/edit the svg with other tools.

I tried to hack this together using set_outer_html(), but this doesn't work for unmounted nodes.

We could do `svg().attr("XMLNS", XMLNS).attr("fill", "currentColor").attr("viewBox", ...).dangerously_set_inner_html("<path ...")`, or something, potentially. Could use [`html_parser::Dom::parse`](https://docs.rs/html_parser/latest/html_parser/struct.Dom.html#method.parse) to strip the `svg` tag from the string, and extract the viewbox.

It might even be possible to make a `svg_view_from_file!("thing.svg")` macro, that does the parsing at compile-time?

What is the preferred approach for sycamore?

noc7c9:
Just posting this because there is activity here. Not sure how relevant/useful this is.

I have a [fork of sycamore](https://github.com/noc7c9/sycamore/commit/6afb6d33daefe7db84f9ecbb7bb3759c051b22a4) where I've implemented (hacked in really) SVG support so that it's usable via the `view!` macro.

```rust
svg(viewBox="-50 -50 100 100") {
    g(style=(format!("transform: rotate({rotation}turn)"))) {
        path(d=(line), fill="none", stroke="red", stroke-width="0.25") {}
        Indexed {
            iterable: points,
            view: move |ctx, (x, y)| view! { ctx,
                circle(cx=(x), cy=(y), r=(RADIUS), fill="red") {}
            },
        }
    }
}
```

This seems to be different to use case being discussed here (dynamic svg vs embedding external svg) but it is what I needed.

[...]

PR #378, opened 2022-03-07
add .inner_html() builder method

alsuren:
This is useful for injecting SVGs into the dom.

I have not tested this with SSR or hydration, because I'm not using them in my current project.

Screenshot 2022-03-07 at 11 56 21


alsuren:
I guess this relates to https://github.com/sycamore-rs/sycamore/issues/288 - we have tried a few things in 0.7-land, like:
* building svgs by hand using the builder api (typically a bad idea, because you end up with an svg that it impossible to view or edit outside of rust-land) 
* falling back to the view!{} macro (okay, but typically forces you to add another layer of div or span that you don't really want).
* encoding the svg as a data url:

	let icon = format!(
		"data:image/svg+xml;charset=UTF-8,{}",
		utf8_percent_encode(source.icon(), ASCII_SET)
	);
        img().attr("src", icon).build()

Since this patch is tiny, we might backport it to 0.7 as well, so I can see how it feels when used with our existing codebase. That way, we give an experience report.

packages/sycamore/src/builder.rs
```diff
@@ -143,6 +143,29 @@ impl<'a, G: GenericNode, F: FnOnce(ScopeRef<'a>) -> G + 'a> ElementBuilder<'a, G
...
+    /// ```compile_fail
+    /// # // This example only fails because we've not supplied a "button_icon.svg" anywhere.
+    /// # // It might be possible to extract some of the SVG constants out of the website dir, and use that here.
+    /// # use sycamore::builder::prelude::*;
+    /// # use sycamore::prelude::*;
+    /// # fn _test<G: GenericNode>(ctx: ScopeRef) -> View<G> {
+    /// h(button).inner_html(include_str!("button_icon.svg"))
+    /// # .view(ctx) }
+    /// ```
+    pub fn inner_html(
```
alsuren, line 162: There's a whole bunch of potential for strange behaviour lurking in here. I can give it a scarier name if you prefer ;-)

lukechu10: I think it should be `dangerously_set_inner_html` to be consistent with the `view!` macro. Not sure if you saw #375 but this PR implements it.

lukechu10: Also it would be somewhat hard to backport to v0.7 because the builder API was completed refactored.

lukechu10:
Sorry haven't taken a look in this in a while. It's that and also I'd like to keep the example the way it is right now to be consistent with the non-builder equivalent. I can push a commit on my side to fix that and merge this soon.

PR #379, opened 2022-03-10
Add index arg to Index component view

shenek:
Before:
```
Indexed {
    iterable: count,
    view: |ctx, x| view! { ctx,
        li { (x) }
    },
}
```

After:
```
Indexed {
    iterable: count,
    view: |ctx, x, idx| {
        let last = count.len() - 1;
        view! { ctx,
            li(class=if last == idx {"active"} else {""}) { (x) }
        }
    },
}
```
## Motivation
It would be nice if index could be directly accessed in view function.
This way it would be easier to e.g. render last component differently or to add extra `class` to odd components, etc.


lukechu10:
I believe that for `Keyed`, `idx` should be a `Signal` instead of just a raw `usize` because this would account for element inserting that would change the indexes.

It would also be nice to have a way to make this optional, so that first of all we don't create unnecessary allocations for the signal, and second of all so that it remains ergonomic for use cases where this is not required. Unfortunately, I don't think this is possible with type trickery, so we'll probably have to create separate components for this.

Perhaps one good reference would be the solidjs implementation (mentioned in the doc comment): https://github.com/solidjs/solid/blob/e7671349791bd239fbe20c54c266ae085a7f09f2/packages/solid/src/reactive/array.ts#L16

Since this increases the scope of this PR, don't feel pressured to work on it if you don't want to. If you do, just continue on this PR. If not, feel free to just close it and somebody else can work on it.

PR #386, opened 2022-03-15
Add path node for svg elements

jafioti:


lukechu10:
Unfortunately, this won't work just like that. This is because SVG elements need to be created in a new namespace otherwise they'll be treated as HTML elements. Also I'd like to add all the SVG elements in one shot rather than just add this one. Sorry about that :(

PR #403, opened 2022-04-04
Allow field init shorthand for component prop fields

danielalvsaaker:
Adds support for field init shorthand syntax, with tests.
 
```rust
let signal = create_signal(cx, 0);
view! { cx, Component { signal } }
```


lukechu10:
This looks great! Thanks!

PR #417, opened 2022-05-07
Handle parsing of component children with function-like syntax

danielalvsaaker:
The parsing logic for components in a view was missing a check for children using function-like calling.

Before this commit, the following snippet would compile:
```rust
view! { cx, Component { Child {} } }
view! { cx, Component { div {} } }
```

...while this would not:
```rust
view! { cx, Component { Child() } }
view! { cx, Component { div() } }
```
<br>
Tests are added to ensure that either way of calling a component compiles.


lukechu10:
Looks great. Thanks!

PR #418, opened 2022-05-07
Add the http-request-builder example

### LINKED ISSUE #395: http-request example with builder syntax
The [http-request example](https://github.com/sycamore-rs/sycamore/tree/master/examples/http-request) already offers information regarding the correct way of fetching data from json APIs. 

Offering the same example via [builder syntax](https://sycamore-rs.netlify.app/docs/basics/view#builder-syntax) would help users that prefer methods instead of macros to better understand both components and Suspense.

mc1098:
This adds the same example as the `http-request` but using the builder
syntax.

Most of the code is the same and a fair degree of the change from macro
to builder syntax is trivial. Using `Suspense` with the builder syntax
does not seem to be covered in the documentation. This also shows off that structs
that derive `Prop` have public builders which may be required using the
builder syntax - such as with `SuspenseProps`.

I added a comment to the `SuspenseProps` part just because I don't think it is obvious
or documented (on the website) that by deriving `Prop` you get a free typed builder too :)

I know @lukechu10 spoke about changing the API around Suspense in the builder
syntax on Discord - so I'm aware you might have reason to not accept this PR yet (or at all) if
those changes are coming soon.

Closes #395

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "examples/context",
   "examples/counter",
   "examples/http-request",
+  "examples/http-request-builder",
   "examples/hydrate",
   "examples/iteration",
   "examples/js-framework-benchmark",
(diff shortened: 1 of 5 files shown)


lukechu10 (approved):
I do in fact plan to change some details about the builder API but I'm happy to merge this PR now because it might be a while before I actually do anything. Just one small comment, could you please add an entry to `examples/README.md` describing the example? Otherwise, LGTM

mc1098:
> could you please add an entry to `examples/README.md` describing the example?

Oh forgot to do that :) 
Done!

lukechu10:
Thanks a lot. I just noticed a small nit in the index.html title but I changed it for you since I didn't want to bother you with such a small change.

PR #419, opened 2022-05-10
Fix destructuring async props

### LINKED ISSUE #410: Allow destructuring in the component signature
For components which take single arguments, i.e. a single field in a `Prop` struct, but also for some more, it would be really useful if something like the following would work:

```
#[derive(Prop)]
struct MyProps<'a> {
    value: &'a ReadSignal<i32>,
}

#[component]
fn MyComponent<'a, G: Html>(cx: Scope<'a>, MyProps { value }: MyProps<'a>) -> View<G> {
    view! {
        div(class="my-component") {
            "Value: " (value.get())
        }
    }
}
```

This would reduce boilerplate and could simulate taking multiple arguments while still transparently using the `Prop` struct. At the moment, this code panics.

mc1098:
Fixes #410

This PR fixes support in the `#[component]` macro for destructured struct patterns of props.

I added a single test to trybuild using the existing prop struct with generics and lifetimes to show it handles those too :)

I started trying to do what I think was [discussed here](https://github.com/sycamore-rs/sycamore/issues/410#issuecomment-1101893839) only to find that trying to match on fields that themselves might also be destructured was a rabbit hole I didn't want to go down :)

So instead this change simply alters the new sync function that we create to wrap around the async one - it takes the type from the destructured pattern and converts it into a `syn::PatType` (i.e. `props: MyProps`) then we just have to add this ident to the args we supply the async function and the pattern the user uses will work if valid and all the errors associated with incorrect patterns will be well formed from rustc.

I added the more eager parser check for the first argument to be `syn::Ident` instead of the unwrap that was there before as this should give a better error message that will be consistent across both types of components.

Moved most of the change to its own function as it got bit messy and used a struct for named fields so it is hopefully more clear :)

### DIFF
--- a/packages/sycamore-macro/tests/view/component-pass.rs
+++ b/packages/sycamore-macro/tests/view/component-pass.rs
@@ -12,6 +12,13 @@ pub fn PropComponent<G: Html>(cx: Scope, Prop { prop: _ }: Prop) -> View<G> {
     }
 }
 
+#[component]
+async fn AsyncPropComponent<G: Html>(cx: Scope<'_>, Prop { prop: _ }: Prop) -> View<G> {
+    view! { cx,
+        div
+    }
+}
+
 #[derive(Prop)]
 pub struct PropWithChildren<'a, G: GenericNode> {
     children: Children<'a, G>,
--- a/packages/sycamore-macro/tests/view/component-pass.rs
+++ b/packages/sycamore-macro/tests/view/component-pass.rs
@@ -12,13 +12,6 @@ pub fn PropComponent<G: Html>(cx: Scope, Prop { prop: _ }: Prop) -> View<G> {
     }
 }
 
-#[component]
-async fn AsyncPropComponent<G: Html>(cx: Scope<'_>, Prop { prop: _ }: Prop) -> View<G> {
-    view! { cx,
-        div
-    }
-}
-
 #[derive(Prop)]
 pub struct PropWithChildren<'a, G: GenericNode> {
     children: Children<'a, G>,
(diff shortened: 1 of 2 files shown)


danielalvsaaker:
This looks very nice, but I think you're doing more work than necessary. [Here's](https://github.com/danielalvsaaker/sycamore/commit/0315d23a1426139979b548a9360de6a1c336426a) my proposed solution. It's okay to reuse the destructured struct as input to the inner async function.

packages/sycamore-macro/tests/view/component-pass.rs
```diff
@@ -22,6 +22,14 @@ pub fn ComponentWithChildren<'a, G: Html>(cx: Scope<'a>, prop: PropWithChildren<
     prop.children.call(cx)
 }
 
+#[component]
+pub fn AsyncComponentWithPropDestructuring<'a, G: Html>(
```
danielalvsaaker, line 26: This function should be async.

danielalvsaaker (changes requested):

mc1098:
> ~This looks very nice, but I think you're doing more work than necessary. [Here's](https://github.com/danielalvsaaker/sycamore/commit/0315d23a1426139979b548a9360de6a1c336426a) my proposed solution. It's okay to reuse the destructured struct as input to the inner async function.~
> 
> Actually, ignore the previous statement. It won't handle partially destructured assignments. Your solution looks good!

Yeah, I don't know how likely someone will use partial destructured patterns for props but good to include it if we can :)

Also, I'm sorry! I didn't realise you were working on the issue and now I feel like I've stepped on your toes - if you want I can close this PR and allow you to continue working on the issue.

lukechu10:
Looks great thanks!

PR #421, opened 2022-05-17
Use create_ref instead of on_cleanup for allocating event handlers

lukechu10:
Also simplifies the implementation for `DomNode::event`.


lukechu10:
Ok the benchmarks are basically useless with this noise.

PR #424, opened 2022-05-20
Fix `scope_depth` to return 0 for root scope

mc1098:
Hi 👋,

Just a quick fix so that the `scope_depth` function returns `0` for the root scope as stated by the comment - 
I also added some tests for the function too.

This function is only ever used in this repository for the following SSR test: https://github.com/sycamore-rs/sycamore/blob/487877f2aee232b4a04c0dd6589a3b2240d292d3/packages/sycamore/tests/ssr/main.rs#L120-L136 
that relies solely on the fact that depth is incremented on child scopes.

### DIFF
--- a/packages/sycamore-reactive/src/context.rs
+++ b/packages/sycamore-reactive/src/context.rs
@@ -113,6 +113,7 @@ pub fn scope_depth(cx: Scope) -> u32 {
     let mut depth = 0;
     let mut current = cx.raw;
 
+    // SAFETY: 'current.parent' necessarily lives longer than 'current'.
     while let Some(next) = current.parent.map(|x| unsafe { &*x }) {
         current = next;
         depth += 1;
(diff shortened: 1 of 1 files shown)


packages/sycamore-reactive/src/context.rs
```diff
@@ -110,11 +110,17 @@ where
...
+    }
+
+    let mut depth = 0;
+    let mut next = cx.raw.parent.map(unchecked_as_ref);
+
+    while let Some(current) = next {
+        next = current.parent.map(unchecked_as_ref);
         depth += 1;
     }
     depth
```
lukechu10, line 126: I have another idea of an implementation that might be simpler:
```rust
while let Some(next) = current.parent.map(unsafe { &*x }) { ... }
```

lukechu10 (changes requested):
I left a few comments. Also thanks for adding tests!

lukechu10 (approved):

PR #425, opened 2022-05-22
Remove `builder` feature doc on website

mc1098:
The `builder` feature flag was removed in #416 but the website still
contained a sentence about adding the feature flag before using the
builder syntax.


lukechu10:
Thanks!

PR #428, opened 2022-05-29
Fix removing old nodes from parent

### LINKED ISSUE #427: TodoMVC example doubles items when switching between All and Completed
**Describe the bug**
When you run the latest version of the TodoMVC example, add some items and then mark some as completed, then when you switch between the tabs All and Completed, the amount of  not completed items will double.

Further, when switching to the completed tab, also uncompleted items will show up. 

**To Reproduce**

1. Run TodoMVC example
2. Add items
3. Mark some as done
4. Switch between All and Completed

**Expected behavior**
I expect the amount of items not to double

**Screenshots**
image

**Environment**

- Sycamore: master
- Browser: Firefox 100.0-1
- OS: Arch Linux

mc1098:
Fixes #427 

The #427 bug was introduced in #416 inside the `reconcile_fragments` function.

The quick fix was to add the negation that got dropped during that change - this seems
to have caused the bug.

~I instead extracted the removal of nodes from the parent node part of the 
`reconcile_fragments` function to a new function and added tests for this new 
function to avoid any future regressions.~

~The tests use a pretty rough/bare bones mock of a `GenericNode` in order to satisfy the
trait bounds required by the extracted function. I think the tests are worth the added file
noise of the mock but let me know if you'd prefer I remove it (and probably the extracted
function at that point).~

### DIFF
--- a/packages/sycamore-core/src/render.rs
+++ b/packages/sycamore-core/src/render.rs
@@ -290,7 +290,7 @@ pub fn reconcile_fragments<G: GenericNode>(parent: &G, a: &mut [G], b: &[G]) {
         } else if b_end == b_start {
             // Remove.
             for node in &a[a_start..a_end] {
-                if map.is_none() || map.as_ref().unwrap().contains_key(node) {
+                if map.is_none() || !map.as_ref().unwrap().contains_key(node) {
                     parent.remove_child(node);
                 }
             }


lukechu10:
Wow that was fast. Thanks! will review soon

packages/sycamore-core/src/render.rs
```diff
@@ -374,3 +370,219 @@ pub fn reconcile_fragments<G: GenericNode>(parent: &G, a: &mut [G], b: &[G]) {
...
+    a_slice: &[G],
+    to_keep: &Option<AHashMap<G, usize>>,
+    parent: &G,
+) {
+    for node in a_slice {
+        if !to_keep
+            .as_ref()
+            .map(|map| map.contains_key(node))
+            .unwrap_or_default()
+        {
```
lukechu10, line 384: Using a combination of `!` and `unwrap_or_default()` really makes this even more confusing (although it was not great to begin with).
```suggestion
        if to_keep.is_none() || !to_keep.as_ref().unwrap().contains_key(node) {
```

mc1098:
Oh the empty mock functions are causing the decrease in coverage 😞

packages/sycamore-core/src/render.rs
```diff
@@ -374,3 +370,219 @@ pub fn reconcile_fragments<G: GenericNode>(parent: &G, a: &mut [G], b: &[G]) {
         }
     }
 }
+
+fn remove_old_nodes<G: GenericNode>(
```
lukechu10, line 374: It would be nice to have a doc-comment explaining what this function does, but it can be really short since it's not public.

lukechu10 (commented):
I think instead of creating a new `MockNode`, it would be cleaner to just re-use `SsrNode` which is what most other tests do. Since `sycamore` depends on `sycamore-core`, we can't add `sycamore` as a dependency but we can add it as a dev-dependency. Unfortunately, rust-analyzer doesn't seem to recognize that but at least it compiles fine.

mc1098:
> I think instead of creating a new `MockNode`, it would be cleaner to just re-use `SsrNode` which is what most other tests do. Since `sycamore` depends on `sycamore-core`, we can't add `sycamore` as a dependency but we can add it as a dev-dependency. Unfortunately, rust-analyzer doesn't seem to recognize that but at least it compiles fine.

Cleaner yes, but I worry about adding point of failures to a unit test that have nothing to do with the function under test - this test may fail because of changes made in the `sycamore-web` which seems wrong imo.

Also I seem to have trouble with compiling the test as the `GenericNode` bound in the new function is not matching the `generic_node::GenericNode` used with the `SsrNode` - I have tried adding the `ssr` feature to the `sycamore` crate dev-dependency which already exists and adding the `sycamore-web` crate (with feature) directly but no dice (this is rustc rejecting it and not just rust-analyzer for me).

The other points I've done locally so no issue there :)

lukechu10:
> Cleaner yes, but I worry about adding point of failures to a unit test that have nothing to do with the function under test - this test may fail because of changes made in the `sycamore-web` which seems wrong imo.

Yeah that's a valid concern but since none of the methods for `GenericNode` are actually called (since all the methods in `MockNode` are basically stubs), changes to `sycamore-web` should affect this. As for the error, that's a bit weird. I'll try to take a look soon.

mc1098:
If you prefer, I can change this PR to do the simple bug fix (add the `!`) and maybe just add tests on the extracted if condition, if you want the regression check retained.

The motivation for the mocking was as a future base for further testing the `reconcile_fragments` function, and possibly others, but this can be moved into another PR that may tackle the refactoring of that function entirely instead.

lukechu10:
> If you prefer, I can change this PR to do the simple bug fix (add the `!`) and maybe just add tests on the extracted if condition, if you want the regression check retained.
> 
> The motivation for the mocking was as a future base for further testing the `reconcile_fragments` function, and possibly others, but this can be moved into another PR that may tackle the refactoring of that function entirely instead.

Alright that sounds like a good idea. Also there are some tests for `reconcile_fragments` already in the `ssr` integration tests (`packages/sycamore/tests/ssr/reconcile.rs`) if you want to continue working on this.

PR #431, opened 2022-06-08
Fix default Unit property on Component

### LINKED ISSUE #415: E0061 when trying to call a component without props in Builder API
**Describe the bug**
I get the Rust error `this function takes 2 arguments but 1 argument was supplied` when trying to call a component without props with the Builder API.

**To Reproduce**
```rust
#[component]
fn App<G: Html>(cx: Scope) -> View<G> {
	// ...
}
#[wasm_bindgen(start)]
pub fn start() {
	sycamore::render(|cx| App(cx)); // ERROR
	sycamore::render(|cx| App(cx, ())); // WORKS
}
```

**Expected behavior**
I expect to be able to call the component like a normal function, as I was able to in the previous version.

**Environment**
- Sycamore: `0.8.0-beta.5`

mc1098:
Removes the default `()` unit property when no properties are
specified in the Component definition.

Removes the above assumption in the `view` macro codegen.
Removes the `UnitBuilder` struct and Unit prop impl from `sycamore-core`.

Fixes #415

### DIFF
--- a/examples/hello-builder/src/main.rs
+++ b/examples/hello-builder/src/main.rs
@@ -26,5 +26,5 @@ fn main() {
     console_error_panic_hook::set_once();
     console_log::init_with_level(log::Level::Debug).unwrap();
 
-    sycamore::render(|cx| component(|| App(cx, ())));
+    sycamore::render(|cx| component(|| App(cx)));
 }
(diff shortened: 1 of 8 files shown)


lukechu10 (approved):

lukechu10:
Looks great thanks! Not sure why I did that in the first place but it's definitely cleaner now.

PR #432, opened 2022-06-08
Add a generic `from_web_sys` function

### LINKED ISSUE #261: Need to be able to convert web_sys::Element to Template<G>
This is important for interfacing with many JS libraries which create arbitrary elements or produce template strings. I was unable to implement `IntoTemplate<G>` due to the blanket implementation, so used instead the `From` trait. It should also be possible in the future to create the template node without needing a wrapper element, though could not find a way from reading through the source. Any tips would be appreciated.

mc1098:
Adds a new function in `sycamore-web` that creates a 
`GenericNode` from a `web_sys::Element` and depending 
on the type will reuse the element in the DOM.

This approach tries to cover the concern of [recreating
existing node](https://github.com/sycamore-rs/sycamore/issues/261#issuecomment-925476447) and returning a `G: GenericNode` that can be used easily.

The function accepts `web_sys::Element` instead of
`web_sys::Node` to avoid having to do error checking
for the case of some non element node. If the user
is using this function then they are probably already
in the know about how to use `JsCast` to get this type
and they can deal with whether they need the overhead
of type checking (in js land anyway).

Not a big fan of the unsafe transmutes in general but 
going the `&dyn Any` downcast route will probably 
effect performance and I think after checking the 
`TypeId` it should be sound. I couldn't really think of
another way of getting over this issue - but if you have
a better way let me know and I'll change it :)

Closes #261

### DIFF
--- a/packages/sycamore-web/src/lib.rs
+++ b/packages/sycamore-web/src/lib.rs
@@ -35,6 +35,40 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
     const IS_BROWSER: bool;
 }
 
+/// Create a `GenericNode` instance from a `web_sys::HtmlElement`.
+///
+/// When the `GenericNode` is a `DomNode` or a `HydrateNode` then the
+/// web_sys::Element will be reused and won't cause another node to be
+/// created in the browser DOM.
+pub fn from_web_sys<G: GenericNode>(element: web_sys::Element) -> G {
+    use std::any::TypeId;
+
+    let type_id = TypeId::of::<G>();
+
+    if TypeId::of::<DomNode>() == type_id {
+        let node = DomNode::from_web_sys(element.into());
+        // SAFETY:
+        // We just made sure that G is a DomNode and we need to transmute the
+        // type as a reference as the reference has the same size
+        let gen_ref: &G = unsafe { std::mem::transmute(&node) };
+        return gen_ref.clone();
+    }
+    #[cfg(feature = "hydrate")]
+    if TypeId::of::<HydrateNode>() == type_id {
+        let node = HydrateNode::from_web_sys(element.into());
+        // SAFETY:
+        // We just made sure that G is a HydrateNode and we need to transmute the
+        // type as a reference as the reference has the same size
+        let gen_ref: &G = unsafe { std::mem::transmute(&node) };
+        return gen_ref.clone();
+    }
+
+    // fall back to inner_html
+    let node = G::element_from_tag("div");
+    node.dangerously_set_inner_html(&element.outer_html());
+    node.first_child().unwrap()
+}
+
 /// Queue up a callback to be executed when the component is mounted.
 ///
 /// If not on `wasm32` target, does nothing.
--- a/packages/sycamore-web/src/lib.rs
+++ b/packages/sycamore-web/src/lib.rs
@@ -16,6 +16,8 @@ mod hydrate_node;
 #[cfg(feature = "ssr")]
 mod ssr_node;
 
+use std::any::{Any, TypeId};
+
 pub use dom_node::*;
 #[cfg(feature = "hydrate")]
 pub use hydrate_node::*;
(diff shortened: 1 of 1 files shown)


packages/sycamore-web/src/lib.rs
```diff
@@ -35,6 +35,40 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
     const IS_BROWSER: bool;
 }
 
+/// Create a `GenericNode` instance from a `web_sys::HtmlElement`.
+///
+/// When the `GenericNode` is a `DomNode` or a `HydrateNode` then the
+/// web_sys::Element will be reused and won't cause another node to be
+/// created in the browser DOM.
+pub fn from_web_sys<G: GenericNode>(element: web_sys::Element) -> G {
```
lukechu10, line 43:
```suggestion
pub fn from_web_sys<G: Html>(element: web_sys::Element) -> G {
```
since this probably shouldn't work for backends such as native or TUI if, and once we have them

packages/sycamore-web/src/lib.rs
```diff
@@ -35,6 +35,40 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
...
+        // We just made sure that G is a HydrateNode and we need to transmute the
+        // type as a reference as the reference has the same size
+        let gen_ref: &G = unsafe { std::mem::transmute(&node) };
+        return gen_ref.clone();
+    }
+
+    // fall back to inner_html
+    let node = G::element_from_tag("div");
+    node.dangerously_set_inner_html(&element.outer_html());
+    node.first_child().unwrap()
```
lukechu10, line 69: I think we can just `panic!` here since if we're not on a web target, we wouldn't have access to a `web_sys::Node` in the first place.

packages/sycamore-web/src/lib.rs
lukechu10, line 43: Also I think the param can just be `web_sys::Node` since this would also work with text nodes and comment nodes. You'll also need to update the doc comment as well.

lukechu10 (changes requested):
Thanks for tackling this. I left a few comments.

lukechu10:
I don't think `&dyn Any` would introduce any performance penalty here since the function will be monomorphisized and the compiler can inline the downcast and then trivially optimize it out. That way, we can avoid the unsafe here.

packages/sycamore-web/src/lib.rs
mc1098: What about `SsrNode`?

lukechu10: on the server side, constructing a `web_sys::Node` would panic already so that should never happen

mc1098:
> Thanks for tackling this. I left a few comments.

No problem just trying to tackling the milestone one step at a time where and while I can 🙂

packages/sycamore-web/src/lib.rs
```diff
@@ -35,6 +35,28 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
...
+        let node = DomNode::from_web_sys(node);
+        return (&node as &dyn Any).downcast_ref().cloned().unwrap();
+    }
+    #[cfg(feature = "hydrate")]
+    if TypeId::of::<HydrateNode>() == type_id {
+        let node = HydrateNode::from_web_sys(node);
+        return (&node as &dyn Any).downcast_ref().cloned().unwrap();
+    }
+
+    panic!("Expected GenericNode to either be a DomNode or a HydrateNode");
```
lukechu10, line 57: nit:
```suggestion
    panic!("expected GenericNode to either be a DomNode or a HydrateNode");
```
(for consistency)

packages/sycamore-web/src/lib.rs
```diff
@@ -35,6 +35,28 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
     const IS_BROWSER: bool;
 }
 
+/// Create a generic `Html` node from a `web_sys::Node`.
+///
+/// # Panics
+/// When G is not either a `DomNode` or a `HydrateNode`.
+pub fn from_web_sys<G: Html>(node: web_sys::Node) -> G {
+    use std::any::{Any, TypeId};
```
lukechu10, line 43: nit: can you please move these to the top of the file?

PR #434, opened 2022-06-11
Add support for `struct` `Route`s

### LINKED ISSUE #244: Support implementing `Route` on `struct`s
Currently, Sycamore explicitly states that the `Route` trait is for `enum`s, and as such its method `match_route` doesn't take the `self` parameter. However, there are use-cases (such as [Perseus](https://github.com/arctic-hen7/perseus)) for which routes may be stored in a `struct`, and can only be accessed through the `self` parameter. I've explored global variables and contexts as possible solutions to this problem, but the former ends up requiring generation of core types and the latter becomes quickly infeasible due to the nature of reactive scopes.

Hence, I would propose that the `Route` trait's methods take the `self` parameter so that they can be feasibly implemented on `struct`s as well as `enum`s. In theory, I would think this should be relatively simple, especially because the rest of the user-facing interface should be able to stay exactly the same, though some changes would be needed to the derive macro for `Route`.

arctic-hen7:
This solves a problem I had with Sycamore last year: that `Route` can only be an `enum`, meaning additional custom information can't be easily attached to the router system. Perseus required this, and did a very hacky workaround that both reduced routing performance there and made for some really weird semantics. With arctic-hen7/perseus#151 though, that workaround it no longer feasible, so this makes `Route` be able to be a `struct`, and `match_route` now takes `&self`.

To achieve this, I've broken out the `Router` into a `Router` and a `RouterBase`, with the former being a wrapper around the latter that simply calls it with `R::default()`, which is now derived automatically in `enum`s as the `#[not_found]` route. All tests pass, and I'm pretty sure I've covered everything, but a review to make sure I haven't missed something would be greatly appreciated!

I haven't done anything with `StaticRouter`, which already takes a specific `Route` instance. `RouterBase` now mimics that, and `Router` just abstracts over it (meaning this isn't a breaking change).

Closes #244.

### DIFF
--- a/packages/sycamore-router-macro/src/route.rs
+++ b/packages/sycamore-router-macro/src/route.rs
@@ -90,7 +90,6 @@ pub fn route_impl(input: DeriveInput) -> syn::Result<TokenStream> {
                 ));
             }
 
-            // We implement `Default` as well here for the `Router`/`RouterBase` distinction (`Router` needs to pass a default `impl Route` to `RouterBase`)
             Ok(quote! {
                 impl ::sycamore_router::Route for #ty_name {
                     fn match_route(&self, __segments: &[&str]) -> Self {
@@ -98,6 +97,7 @@ pub fn route_impl(input: DeriveInput) -> syn::Result<TokenStream> {
                         #err_quoted
                     }
                 }
+                // We implement `Default` as well here for the `Router`/`RouterBase` distinction (`Router` needs to pass a default `impl Route` to `RouterBase`)
                 impl ::std::default::Default for #ty_name {
                     fn default() -> Self {
                         #error_handler_name
--- a/packages/sycamore-router-macro/src/route.rs
+++ b/packages/sycamore-router-macro/src/route.rs
@@ -98,7 +98,7 @@ pub fn route_impl(input: DeriveInput) -> syn::Result<TokenStream> {
                         #err_quoted
                     }
                 }
-                impl Default for #ty_name {
+                impl ::std::default::Default for #ty_name {
                     fn default() -> Self {
                         #error_handler_name
                     }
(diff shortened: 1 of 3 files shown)


arctic-hen7:
Note that only `Clone`able data will be able to be stored in the `Route` right now, since it takes `&self` rather than `self`. if it takes the latter though, that stuffs up some lifetimes in another part of the code, which I'm not completely sure how to quickly fix. For my purposes, this is fine, but it might be worth revisiting in future.

packages/sycamore-router-macro/src/route.rs
```diff
@@ -82,20 +83,26 @@ pub fn route_impl(input: DeriveInput) -> syn::Result<TokenStream> {
...
 
-            if !has_error_handler {
+            if error_handler_name.is_none() {
                 return Err(syn::Error::new(
                     input.span(),
                     "not found route not specified",
                 ));
             }
 
+            // We implement `Default` as well here for the `Router`/`RouterBase` distinction (`Router` needs to pass a default `impl Route` to `RouterBase`)
```
lukechu10, line 93: Can you move this comment next to where the impl block is? It's fine if it's in the `quote!` because it is ignored by the macro.

lukechu10 (commented):
Sorry don't have time for a full review right now. I'll try to get to this soon.

arctic-hen7:
No problem, Perseus can run off my fork in the meantime.

packages/sycamore-router-macro/src/route.rs
arctic-hen7: Done.

arctic-hen7:
@lukechu10, would it be possible to merge this fairly soon and get a new beta release out? I'd like to make a new release of Perseus fairly soon to field test the removal of `.perseus/`, which depends on this.

lukechu10:
Sorry for taking so long to get to this. I'm planning to do a refactor of the router API for the final 0.8 release but I'll merge this in for now to unblock you.

lukechu10:
Released v0.8.0-beta.7!

PR #435, opened 2022-06-11
Add the `prop:*` directive to the view macro

### LINKED ISSUE #330: Set element property directly from view! macro
The view macro allows setting elements attributes (ex: class, id, style, etc) but does not allow setting or binding to element properties. 

While most HTML element properties and attributes have a 1-to-1 relationship (the attribute is the HTML serialisation of the underlying properties state) -- this is not always the case. Many properties do not have associated attributes. 

One single example of this is the `indeterminate` property on `<input type="checkbox"/>`. 

This property cannot be set or checked by using html attributes.

mc1098:
Adds the ability to set properties directly in the view macro that
accepts any value that can be converted into the `JsValue`.

Supports both static and dynamic values in a similar manner that
attributes do currently.

Updates some of the examples to use the new directive when the type can
be used directly as a property instead of cloning or serializing the
value.

I did update the website to include a section on this new directive but there
is not a similar section for the builder API, so I'm not sure if you want to keep
this out or whether you'd like me to add something similar for the builder docs :).

Closes #330

### DIFF
--- a/docs/next/basics/view.md
+++ b/docs/next/basics/view.md
@@ -118,6 +118,24 @@ view! { cx,
 
 Instead, when displaying user input, use interpolation syntax instead.
 
+### Properties
+
+Properties are set using the `prop:*` directive.
+
+```rust
+view! { cx,
+    input(type="checkbox", prop:indeterminate=true)
+}
+```
+
+There are some properties that do not have an attribute, such as
+`indeterminate` in HTML, which must be set using the `prop:*` directive.
+
+There are a number of properties that have an associated attribute, such as
+`value`, in these cases an attribute is deserialized to become the state of the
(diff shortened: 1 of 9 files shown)


mc1098:
Lets see if I can make Codecov happy this time :)

lukechu10:
Codecov is indeed happy this time but don't worry too much about it. The reason is that half of our test suite does not actually count towards coverage because they run only in wasm. In fact, our coverage should be much better than what it shows.

lukechu10 (approved):

lukechu10:
Sorry for taking so long to review. LGTM. As for the docs, I'm planning on rehauling it anyways, so I'll handle that.

PR #436, opened 2022-06-12
Add doc test example for `from_web_sys`

mc1098:
_if someone wants to contribute a doctest for this function, that will be very welcome!_

I should have probably have added it in the initial PR tbh :)

### DIFF
--- a/packages/sycamore-web/src/lib.rs
+++ b/packages/sycamore-web/src/lib.rs
@@ -49,7 +49,6 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
 /// #     todo!()
 /// # }
 /// use sycamore::prelude::*;
-///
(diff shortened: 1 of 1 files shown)


packages/sycamore-web/src/lib.rs
```diff
@@ -41,6 +41,27 @@ pub trait Html: GenericNode<EventType = Event, PropertyType = JsValue> {
...
 /// When G is not either a `DomNode` or a `HydrateNode`.
+///
+/// # Examples
+///
+/// ```
+/// # fn create_raw_node() -> web_sys::Node {
+/// #     todo!()
+/// # }
+/// use sycamore::prelude::*;
+///
```
lukechu10, line 52:
```suggestion
```
Otherwise there would be an extra space in the rendered docs.

PR #438, opened 2022-06-20
Add `trigger_subscribers` to `Signal`

mc1098:
Exposes `trigger_subscribers` function from `SignalEmitter` to `Signal`'s public interface.

### DIFF
--- a/packages/sycamore-reactive/src/signal.rs
+++ b/packages/sycamore-reactive/src/signal.rs
@@ -189,7 +189,7 @@ impl<T> Signal<T> {
     /// ```
     pub fn set(&self, value: T) {
         self.set_silent(value);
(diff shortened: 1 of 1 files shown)


lukechu10 (changes requested):
Since we're adding this, would you mind changing all the other instances of `self.0.emitter.trigger_subscribers()` (in `Signal::set` and `Signal::set_rc`) to just `self.trigger_subscribers()`? Thanks for contributing!

mc1098:
> Since we're adding this, would you mind changing all the other instances of `self.0.emitter.trigger_subscribers()` (in `Signal::set` and `Signal::set_rc`) to just `self.trigger_subscribers()`? Thanks for contributing!

Done :)

lukechu10:
:tada:

PR #440, opened 2022-06-21
Update boolean attributes in the view macro

mc1098:
This adds some attributes but mostly removes attributes which are not
part of the standard.

I have gone through each attribute removed in this list just to make sure I'm
not removing something often used but not in the WHATWG standard for some
reason - this was never the case.

I can, if you like, explicitly list each removed attribute and link to some evidence
that it is either (1) not a boolean attribute (like `contenteditable`); (2) deprecated
(like `scoped`); or (3) is not going to be implemented (like `sortable`).

As is this PR would be in sync with the boolean attributes listed in the [WHATWG
attributes table](https://html.spec.whatwg.org/multipage/indices.html#attributes-3).

This wouldn't prevent a user from using one of the removed attributes as a boolean
attribute in the case they are targeting some old version of a browser. They can
use the attribute name and set the value to either "<name of attribute>" or to an empty
string but to remove it would probably need `NodeRef` and manually calling to
remove the attribute.


lukechu10:
Awesome! I forgot where I got the original list of boolean attributes, but it seems like it was quite incomplete/outdated!

PR #441, opened 2022-06-23
Add debug implementations

### LINKED ISSUE #412: Implement `Debug` for public `struct`s
In my experience, it's generally a good idea to do this, even if the implementation just says the name of the `struct`. Also, I need to put `Scope`s into `struct`s that implement `Debug` for Perseus, so it would be nice if `BoundedScope` implemented it.

I'm happy to work on a PR for this that just implements fairly naive `Debug`s (if it can't be derived, just say the name of the `struct` and move on) on all public `struct`s, but I thought I should open this first to check there aren't any objections.

arctic-hen7:
This adds `Debug` implementations to all public items exported by all Sycamore crates, additionally adding the restriction that this trait must be implemented for all future exports.

Closes #412.


lukechu10 (approved):
LGTM. Thanks!

PR #442, opened 2022-06-26
Add debug impl for RouterBaseProps

mc1098:
Missing debug after #441 added by #434.

I noticed this because the website failed to build.


lukechu10:
Thanks!

PR #449, opened 2022-07-01
Refactor web tests

mc1098:
Hi 👋,

This PR does five things for the tests in `packages/sycamore/tests/web/`. 

1. Add `query` utility function to replace `{document}.query_selector({selectors}).unwrap().unwrap()`
2. Add `query_into` utility function to replace `{document}.query_selector({selectors}).unwrap().unwrap().unchecked_into::<{element_type}>()`
3. Add a new `assert_text_content` macro, which is similar to `assert_eq` but it will call `Node::text_content` and `unwrap` the value for the `left` expression before passing it to `assert_eq`.
4. Uses the methods available in `HtmlInputElement` for getting and setting the `value` property instead of using the `Reflect` methods.
5. I noticed the svg test which didn't assert that the class was added correctly, so I thought I'd add the assertion for this :)

Hopefully you agree this makes writing and reading these types of tests easier :)

### DIFF
--- a/packages/sycamore/tests/web/builder_hydrate.rs
+++ b/packages/sycamore/tests/web/builder_hydrate.rs
@@ -132,35 +132,14 @@ mod dynamic {
 
             sycamore::hydrate_to(|_| v(cx, state), &c);
 
-            assert_eq!(
-                c.query_selector("p")
-                    .unwrap()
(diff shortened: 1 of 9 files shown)


lukechu10 (approved):
Wow this is so much better than before!

PR #475, opened 2022-08-31
Forward compatibility fix for breakage due to a soundness fix in rustc

oli-obk:
In https://github.com/rust-lang/rust/pull/95474#issuecomment-1228705007 we are attempting to deploy a soundness bug fix for rare cases of `impl Trait` causing UB. While your crate does not use this unsoundly, it requires an additional annotation in order for rustc to be able to figure out that your case is actually sound.

This change was done by following the [diagnostics emitted by the rustc that contains the bugfix](https://crater-reports.s3.amazonaws.com/pr-95474-1/try%23d47ae2a6ecee02f32f23667fe1152f9a71e39e8c/reg/perseus-0.4.0-beta.7/log.txt).


lukechu10:
Thanks a lot! I am always impressed at how proactive the Rust community is!

Issue #468, opened 2022-08-19
`bind_value` doesn't accept `RcSignal` value


zaphar:
In the ElementBuilder API the `bind_value` method doesn't take `RcSignal` values. This signal type is occasionally necessary to work around scoping issues with the context. It would be nice if that method could take either a Signal or an RcSignal.

Using Sycamore: [e.g. 0.8.0-beta.7, master]

danielalvsaaker:
Since `RcSignal<T>` implements `Deref<Target = Signal<T>>`, can't you just pass a reference to your `RcSignal` value?

zaphar:
I have not tried that workaround yet. I'm going to take another stab at moving onto v0.8 soon so I'll give it a try then.

PR #489, opened 2022-09-22
Allow any signal to be bound

danielalvsaaker:
Suggestion on how to make binding signals more ergonomic, in reference to #468.

This will clone the input value if it's a `RcSignal`, or in practice perform a copy if it's a `&Signal`.


danielalvsaaker:
Sure, I'll have a go at that in this PR

packages/sycamore-macro/src/view/codegen.rs
```diff
@@ -433,16 +433,20 @@ impl Codegen {
...
                     });
                     ::sycamore::generic_node::GenericNode::event(&__el, #cx, #event_name,
-                        ::std::boxed::Box::new(|event: ::sycamore::rt::Event| {
+                    {
+                        let #expr = ::std::clone::Clone::clone(&#expr);
+                        ::std::boxed::Box::new(move |event: ::sycamore::rt::Event| {
                             #expr.set(#convert_from_jsvalue_fn);
-                        }),
+                        })
+                    },
```
danielalvsaaker, line 449: Introduces implicit cloning of signals in the same way as for the builder API.

This solves issues like this:
```rust
let signal = create_rc_signal(String::new());
view! { cx,
    input(bind:value=signal)
    (signal.get())
}
```

Note that order will matter. Binding a signal after reading it will still give a "value borrowed after move" error.

PR #491, opened 2022-09-23
Add proper error handling on the http-request example.

ar37-rs:


ar37-rs:
Hi @lukechu10, i've decided to simplify and it turned out way more simpler and working as expected.

Issue #501, opened 2022-10-11
r#use not resolving to correct <use> html tag


danielnehrig:
**Describe the bug**
using r#use does not resolve to <use> tag

**To Reproduce**
```rust
view! {cx,
  svg() {
    r#use()
  }
}
```

**Expected behavior**
r#use does produce a proper <use> tag

**Screenshots**
<img width="163" alt="Screenshot 2022-10-11 at 2 06 54 AM" src="https://user-images.githubusercontent.com/4050749/194970321-915f9af9-4f51-45c3-8567-7c68c5009f7d.png">

**Environment**

- Sycamore: 0.8.2
- Browser: chrome
- OS: mac

danielnehrig:
https://github.com/sycamore-rs/sycamore/pull/502/files

Issue #500, opened 2022-10-11
linearGradient hydration error


danielnehrig:
**Describe the bug**
hydration error expected `linearGradient` found `lineargradient`

```
bundle.js:342 panicked at 'hydration error, mismatched element tag
expected linearGradient, found lineargradient
occurred at element with hydration key 3.27', /Users/dNehrig/.cargo/registry/src/github.com-1ecc6299db9ec823/sycamore-web-0.8.2/src/hydrate_node.rs:102:17

Stack:

Error
    at imports.wbg.__wbg_new_abda76e883ba8a5f (http://localhost:8008/.perseus/bundle.js:345:17)
    at console_error_panic_hook::Error::new::hfd696b293d5ca8a1 (http://localhost:8008/.perseus/bundle.wasm:wasm-function[13018]:0x49fdd9)
    at console_error_panic_hook::hook_impl::h63875983ce5287cc (http://localhost:8008/.perseus/bundle.wasm:wasm-function[2526]:0x2f660b)
    at console_error_panic_hook::hook::hf081cf4b2f9853fb (http://localhost:8008/.perseus/bundle.wasm:wasm-function[14406]:0x4b298e)
```

**To Reproduce**
import this component into your sycamore app load the page and look into the console
```rust
use perseus::Html;
use sycamore::prelude::*;

#[component]
pub fn Ferris<G: Html>(cx: Scope) -> View<G> {
    view! { cx,
        svg(xmlns="http://www.w3.org/2000/svg", width="385", height="256", viewBox="0 0 1200 800") {
            g(id="Layer-1", serif_id="Layer 1") {
                g(transform="matrix(1,0,0,1,597.344,637.02)") {
                    path(d="M0,-279.559C-121.238,-279.559 -231.39,-264.983 -312.939,-241.23L-312.939,-38.329C-231.39,-14.575 -121.238,0 0,0C138.76,0 262.987,-19.092 346.431,-49.186L346.431,-230.37C262.987,-260.465 138.76,-279.559 0,-279.559", style="fill:rgb(165,43,0);fill-rule:nonzero;")
                }
                g(transform="matrix(1,0,0,1,1068.75,575.642)") {
[...]

PR #502, opened 2022-10-13
fix(bug): #500 fix and #501 fix

danielnehrig:
Signed-off-by: Daniel Nehrig <daniel.nehrig@dnehrig.com>


danielnehrig:
bench is failing but i don't think its related to the changes

danielnehrig:
you want me to do here something else or can we merge ?

PR #511, opened 2022-10-18
Add data binding for valueAsNumber property

### LINKED ISSUE #464: Add data binding for `valueAsNumber` property
`valueAsNumber` is like `value`, but it's a number instead of a string. It's useful for sliders.

wainwrightmark:
Closes #464


danielalvsaaker:
Camelcase looks very out of place in my opinion. Could we call it just "number"? `bind:number` looks more right to me.

PR #529, opened 2022-11-11
adding `change` and `change_silent`

blainehansen:
Hey there! :)

I thought a convenience function like `change` would be nice in order to avoid needing a `state.get()` call for simple use cases. This is done in `solidjs` as well: https://www.solidjs.com/tutorial/introduction_signals

Let me know if there are any changes you'd like made :)

P.S., I chose `Fn(&T) -> T` to avoid requiring `Copy` or `Clone` on T. This is only really intended for the simplest situations.

### DIFF
--- a/packages/sycamore-reactive/src/signal.rs
+++ b/packages/sycamore-reactive/src/signal.rs
@@ -228,6 +228,25 @@ impl<T> Signal<T> {
         self.trigger_subscribers();
     }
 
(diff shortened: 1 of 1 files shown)


blainehansen:
I'm not attached to the name :) Perhaps something along the lines of `set_fn` would make more sense, just to align more with `set`?

PR #536, opened 2022-11-30
(Runtime) Templates

### LINKED ISSUE #19: Clone template nodes instead of using document.createElement
For better initial rendering performance, nodes should be created from a template node rather than using `document.createElement`. This would also improve component creation speed.

lukechu10:
Closes #19 

Adds a template system where `Template` represents the structure of the `view!` with holes inside. These holes can then be binded to and filled in with the dynamic values.

This should give a big performance boost on the client side because we are cloning templates now instead of creating each and every element individually.
This could also give a performance boost on server-side by allowing us to cache the SSR-ed strings.

This also unlocks other exciting features such as some form of hot view-reloading.

### DIFF
--- a/packages/sycamore-macro/src/view/codegen.rs
+++ b/packages/sycamore-macro/src/view/codegen.rs
@@ -134,7 +134,7 @@ impl CodegenTemplate {
             .map(|child| self.node(child))
             .collect::<Vec<_>>();
 
-        let ret = match tag {
+        match tag {
             ElementTag::Builtin(tag) => quote! {{
                 type __tag = #elements_mod_path::#tag;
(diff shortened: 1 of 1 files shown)


arctic-hen7:
I'd just like to briefly mention that Perseus also has a `Template` system...is it too late to consider another name for this perhaps to reduce confusion? (This looks fantastic btw!)

arctic-hen7:
Ah understood! Fantastic, thank you very much! (Side note: this might be a great opportunity for some good architectural docs for Sycamore, which I'd be happy to help write. Would you like me to open a new tracking issue for this?)

Issue #200, opened 2021-08-06
Allow on-demand upgrade from static html to js components (islands)


ibotty:
The idea is described here https://jasonformat.com/islands-architecture/

There was a comment on reddit asking for the same thing.
https://www.reddit.com/r/rust/comments/ouqfg3/new_tokio_blog_post_announcing_axum_web_framework/

lukechu10:
This is currently blocked on #81. Right now, the original DOM is destroyed and recreated.

It would be nice if we could provide support for https://github.com/snowpackjs/astro. That should provide quite a bit of heavy lifting for supporting this.

redactedontop:
Updates?

PR #537, opened 2022-11-30
Remove data-hk attribute after hydration

sorpaas:
Currently, hydrating multiple components on the same page in sycamore does not work. This is because, across mount boundaries, elements can have the same `data-hk` value and thus confuses sycamore.

This can be fixed by simply removing `data-hk` immediately after each element is hydrated. We can only do in-order from-top-to-bottom multi-component hydration, but that's more than sufficient.

This PR allows sycamore to [support island architecture](https://github.com/corepaper/east/tree/master/examples/counter). rel https://github.com/sycamore-rs/sycamore/issues/200


lukechu10:
Superseded by #679

PR #548, opened 2023-01-01
Allow passing through attributes on components

wingertge:
This allows a component to take in a map of arguments so they can be passed through to an inner HTML element.
The current syntax proposal is the following:

# Passing attributes

Any prefixed "props" (`on:`, `bind:`) will be automatically passed through. For attributes that don't have a built in prefix, `attr:` can be added to pass it through.

## Example:
The class and on click event will be passed through via `Attributes`, `is_accessible` is a regular prop.
```rust
view! { cx, Label(attr:class = "bg-gray-400", on:click = on_click, is_accessible = true) { "Name" } }
```

# Applying attributes to elements

Applying will be done with Rust's "rest" operator.

## Example:
```rust
view! { cx, label(..attributes, id = aria_id) { (children) } }
```

# Example Use Case
[Accessible Checkbox](https://github.com/wingertge/headlessui-sycamore/blob/master/src/components/checkbox.rs)

# Current status

The code is very suboptimal and documentation will be added when the feature is done. The following parts of the feature are currently done:
- [x] The `Attribute` struct and `AttributeValue` enum required to pass through values in a type safe way
- [x] Parsing prefixed props as attributes and generating the `Attribute` struct from them
- [x] Applying attributes to an element
- [x] Tests
- [x] Cleanup and documentation
- [x] More ergonomic way to get values from `Attributes`
- [x] Example
- [x] Book documentation


wingertge:
I agree with you in disliking the `AttributeValue` enum, I think it's very inelegant, but the perfomance is honestly a very minor reason for it.

The main reason is the error handling story.`Expected &'str, found {number}` is a lot more readable than `Expected 'dyn Display', found {number}`. Of course the dynamic version still has the latter message, but I think keeping the errors as good as possible in as many situations as possible is important.

Making it entirely dynamic (i.e. Box<dyn Any>) would absolutely obliterate the error story and make reading or modifying attribute keys very painful because for all you know the user could've accidentally put in the wrong type in the attribute and now you're having downcast errors at runtime.

As for letting the user do what they want with attributes I think this is fine - it has exactly the same restrictions as the `view!` macro already does. The only difference in behaviour is with boolean, because the view macro only takes booleans directly right now and to get reactivity we need to pass the signal through into the component. Unfortunately `Into<bool>` can't be made into a box, so dynamic boolean values are currently required to be signals.

wingertge:
> I was originally thinking along the lines of using `NodeRef`, perhaps wrapped behind another struct. This way, the codegen for the attributes on a component and on a "plain" element would be almost identical, the former operating on the `NodeRef` and the latter operating on the element directly. Everything else, including the rest operator would still remain the same API wise.
The thing with that approach is that you can't set attributes until the noderef is set, right? Which happens after the view macro for the component is already executed. So you still need some intermediate data structure that stores the values until you get to the point of use. Which would land you in the same place as `AttributeValues` anyways.
> Another possibility would be to make `AttributeValue` dynamic for everything and to get rid of the static variants. On the upside, this would make the code slightly cleaner. However, this would also have slightly worse performance but hopefully should not be noticeable.
I think there should be a user facing API for getting values of certain type, while resolving any dynamic stuff. It would then be up to the user to use that API in a `create_effect` or whatever to make sure they don't lose reactivity. For the implementation I still think the overhead in codegen is relatively minimal compared to the improved error story.

wingertge:
> One way to work around this is to first call the component and therefore have the `NodeRef` set. Then you can attach all the attributes after the component is called. This means however that the component will not be able to inspect the attribute values unless something like `on_mount` is used.

I think that would be a complete non-starter. My example use case of accessible p components needs the ID set on first call so it can pass it through to the children via context. This would not be possible in `on_mount`.

> I think it should be possible to have good error messages even if everything is dynamic. We are already manually checking if attributes should be strings or bools so with some macro magic, error messages should be just as good.

It's possible of course. The question is: why spend so much effort recreating the error messages instead of relying on the compiler when you'll probably end up with more code bloat than `AttributeValues` adds?

> Also note that having `Box<dyn Display>` or similar does not actually make the value "dynamic" in the sense that it is a reactive value. This merely means that anything that can be `Display`ed can be used as an attribute. Instead, if we are to stick with the `AttributeValue` enum route, we could have `Attributes` be a struct containing `Box<dyn Fn() -> AttributeValue>`s instead which will be reactive.

`Box<dyn Display>` is reactive because it can contain a signal and that signal gets tracked when `to_string` is called inside a `create_effect`. That's what the `view!` macro does already, it just stringifies any non literal value in a `create_effect` to make it reactive if it's a signal.

lukechu10:
> As for the issue of overriding attributes a user is not supposed to touch, that's where the API for `Attributes` comes back in.

In this case, should the attributes be silently ignored, or should there be a panic at runtime? I would say the later.

In your example, I think it might be better to only use the prop `id` and panic if there is also an `id` attribute as well. This way, there is only a single way to do things.

wingertge:
> In this case, should the attributes be silently ignored, or should there be a panic at runtime? I would say the later.
> 
> In your example, I think it might be better to only use the prop `id` and panic if there is also an `id` attribute as well. This way, there is only a single way to do things.

That's debatable, I tend to err on the side of silently fixing any mistakes the user has made because it's makes it more convenient to use, but I think that should be up to the library developer anyways and shouldn't be included on `Attributes` itself.

Are there any other issues or should I start working on the more ergonomic API and then start adding docs/an example?

lukechu10:
> That's debatable, I tend to err on the side of silently fixing any mistakes the user has made because it's makes it more convenient to use, but I think that should be up to the library developer anyways and shouldn't be included on `Attributes` itself.

Ok fair enough.

> Are there any other issues or should I start working on the more ergonomic API and then start adding docs/an example?

We can stick with the `AttributeValue` enum for now and we can always iterate or change it later if we need to. I'll do a full code review now.

packages/sycamore-core/src/component.rs
```diff
@@ -176,3 +181,108 @@ impl<'a, G: GenericNode> Children<'a, G> {
...
     }
 }
+
+/// The value of a passthrough attribute.
+/// The default for unknown attributes is [`AttributeValue::Str`] or [`AttributeValue::DynamicStr`]
+pub enum AttributeValue<'cx, G: GenericNode> {
+    /// A string literal value. Example: `attr:id = "test"`
+    Str(&'static str),
+    /// A dynamic string value from a variable. Example: `attr:id = id_signal`
+    DynamicStr(Box<dyn Display>),
```
lukechu10, line 191: I still think it would be clearer to have a `Box<dyn FnMut() -> String>` instead. If you have a `ReadSignal`, you can easily create a closure with `|| signal.get().to_string()`

This way, it is immediately clear that there can be a reactive access inside the closure whereas the `dyn Display` is a lot more obscure.

Same goes for `DynamicBool` and `DynamicDangerouslySetInnerHtml`.

packages/sycamore-core/src/component.rs
```diff
@@ -176,3 +181,108 @@ impl<'a, G: GenericNode> Children<'a, G> {
...
+    }
+}
+
+impl<'cx, G: GenericNode> Deref for Attributes<'cx, G> {
+    type Target = HashMap<Cow<'static, str>, AttributeValue<'cx, G>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.attrs
+    }
+}
```
lukechu10, line 282: I don't know if we should expose the underlying `HashMap` directly. I think it would be better to limit the operations to just insert and remove.

packages/sycamore-core/src/generic_node.rs
```diff
@@ -127,6 +127,17 @@ pub trait GenericNode: fmt::Debug + Clone + PartialEq + Eq + Hash + 'static {
...
+    /// The event should be removed once the scope is disposed, as to prevent accessing scope
+    /// variables after the scope is disposed.
+    /// This is the boxed version in case the handler is already boxed
+    fn boxed_event<'a>(
+        &self,
+        cx: Scope<'a>,
+        name: &str,
+        handler: Box<dyn FnMut(Self::EventType) + 'a>,
+    );
+
```
lukechu10, line 140: I think adding this extra method is a bit unnecessary since the performance overhead of an extra box is negligible anyways.

packages/sycamore-macro/src/view/codegen.rs
```diff
@@ -375,6 +375,169 @@ impl CodegenTemplate {
                 });
                 (None, true)
             }
+            AttributeType::Spread => {
```
lukechu10, line 378: All the code in the `quote!` should be extracted into a method in some utility `mod` in `sycamore`. We probably don't want to emit all this every time we use a spread because that would dramatically bloat the macro output and potentially slow down compile-times.

packages/sycamore-macro/src/view/parse.rs
```diff
@@ -175,43 +177,49 @@ impl Parse for AttributeType {
...
-                "bind" => {
-                    let prop = input.call(Ident::parse_any)?;
-                    Ok(Self::Bind {
-                        prop: prop.to_string(),
-                    })
+        let lookahead = input.lookahead1();
+        if lookahead.peek(Token![..]) {
+            let _2dot = input.parse::<Token![..]>()?;
+            Ok(Self::Spread)
+        } else {
```
lukechu10, line 184: For better error messages, we should also have `else if lookahead.peek(Ident)` and an `else { Err(lookahead.error()) }`

packages/sycamore-macro/src/view/parse.rs
```diff
@@ -248,11 +256,55 @@ impl Parse for Component {
...
     fn parse(input: ParseStream) -> Result<Self> {
-        Ok(Self {
-            name: input.parse()?,
-            eq: input.parse()?,
-            value: input.parse()?,
-        })
+        pub struct AttributeName {
+            tag: Ident,
+            extended: Vec<(Token![-], Ident)>,
+        }
```
lukechu10, line 262: We should probably reuse the `AttributeName` from above to prevent code-duplication, since we are parsing the exact same syntax.

wingertge:
> We can stick with the `AttributeValue` enum for now and we can always iterate or change it later if we need to.

That's part of why I want to add a more ergonomic API too, makes it easier to swap out the backend if people are mostly using `AttributeValue` indirectly through the helper methods on `Attributes`.

packages/sycamore-core/src/component.rs
wingertge: I thought it might be a good idea to have raw access in case you need advanced matching for `AttributeValues`, but I think the ergonomic API should actually entirely replace it because attributes always have a defined type anyways, so the type is known from the key. I'll push a commit now.

packages/sycamore-core/src/component.rs
wingertge: I disagree but it's not a hard change I think so I'll just give it a try

packages/sycamore-core/src/generic_node.rs
wingertge: You're right, I have a tendency to over-optimize when I write code. I removed the method.

packages/sycamore-macro/src/view/codegen.rs
wingertge: You're right, it would've made writing it a lot easier too. Oh well, working on moving it and cleaning it up

wingertge: Done

packages/sycamore-macro/src/view/parse.rs
wingertge: Actually the error message will be perfectly fine, it's the same one it was before. It'll just say "expected Ident, found {blank}".

wingertge: I literally just put the old code in the `else` statement

packages/sycamore-macro/src/view/parse.rs
wingertge: Thanks for reminding me, I wanted to do that before but forgot

wingertge:
@lukechu10 I resolved all your issues with the code, want to give it another lookover?

lukechu10:
> @lukechu10 I resolved all your issues with the code, want to give it another lookover?

Yeah will try to do soon.

wingertge:
Scratch that, I just merged in the new changes to event in an attempt to resolve the merge conflict and I'll need to give some thought to how I'm going to implement the new event syntax with `AttributeValues`. I might need to add back a version of `boxed_event` that lets me pass a dynamic function.

The problem is that `EventHandler` has the descriptor as a generic parameter, which makes it impossible to just box it as dynamic, since the descriptor is also dynamic and can't be set as a parameter.

packages/sycamore-macro/src/view/parse.rs
wingertge: That's a fair point. I'll look into it once I have the event stuff reimplemented.

wingertge:
It wasn't quite as bad as I thought it would be. `event` already type erases it, so I just copied that code into the code for creating `Attributes` and added an untyped version of event. The updated version is ready for review.

packages/sycamore-macro/src/view/parse.rs
wingertge: the `if lookahead.peek(Ident)` version doesn't work because `Ident::parse` includes keywords but `peek(Ident)` does not, which fails on `ref`. I instead used `map_err` on the name parse to show the improved error.

wingertge:
Codecov is complaining now because all the new code from the ergonomic API isn't tested, but I feel like testing simple wrappers around `HashMap::get` and `HashMap::remove` is kind of silly

wingertge:
Since the general API is pretty much done and we're just talking small fixes to the implementation details now, I took the liberty of adding a page to the docs and an example. Don't want to make the existing docs problem even worse by adding an undocumented feature.

packages/sycamore-macro/src/view/parse.rs
lukechu10: Sorry for taking so long to respond. It is possible to do use [`Ident::peek_any`](https://docs.rs/syn/latest/syn/ext/trait.IdentExt.html#associatedconstant.peek_any) which has the analogous behavior to `Ident::parse_any`

wingertge: No worries, I was able to use my time to implement this change on the headless-ui port I mainly designed it for. I can do the peek change if you want, I figured that out after doing the error map but didn't think it was a big enough difference to warrant a commit. I'll get it done now real quick.

lukechu10 (commented):
The end-user API looks great! A few comments here and there but mostly minor things.

lukechu10:
Also don't mind codecov too much. In fact, the actual coverage should be _much_ higher but it isn't because a lot of the tests are run in wasm instead.
I treat is as more of an indicator that some things aren't tested rather than an absolute metric.

docs/next/advanced/attributes_prop.md
```diff
@@ -0,0 +1,57 @@
+# `Passing Attributes to a wrapped element`
```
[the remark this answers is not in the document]
wingertge: I just saw your other pages do it like that so I thought that was important for the styling or something.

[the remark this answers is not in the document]
wingertge: apparently it's just the `NodeRef` page, which just happened to be the exact one I looked at for reference

packages/sycamore-reactive/src/signal.rs
```diff
@@ -566,6 +566,23 @@ impl<T: Display> Display for ReadSignal<T> {
...
+impl From<&Signal<bool>> for bool {
+    fn from(signal: &Signal<bool>) -> bool {
+        *signal.get()
+    }
+}
+impl From<&ReadSignal<bool>> for bool {
+    fn from(signal: &ReadSignal<bool>) -> bool {
+        *signal.get()
+    }
+}
```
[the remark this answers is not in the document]
wingertge: You're right. I thought it was required but I can actually just put the whole expression in a closure and it'll work with the current syntax. I still think it should be added specifically to allow third party code to accept both signals and non-reactive values as a single prop and it also makes sense since `Display` and `Hash` are implemented, but like you said that's a separate PR.

packages/sycamore-web/src/dom_node.rs
```diff
@@ -293,6 +293,22 @@ impl GenericNode for DomNode {
             .unwrap_throw();
     }
 
+    fn untyped_event<'a>(
```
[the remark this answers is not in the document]
wingertge: Good idea

packages/sycamore/src/utils.rs
```diff
@@ -40,3 +46,135 @@ pub fn initial_node<G: GenericNode>(_el: &G) -> Option<View<G>> {
         None
     }
 }
+
+#[doc(hidden)]
+pub fn erase_handler<'a, Ev, Handler, S>(
+    cx: Scope<'a>,
+    mut handler: Handler,
+) -> Box<dyn FnMut(JsValue) + 'static>
```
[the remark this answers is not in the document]
wingertge: I double checked and it does need to be `'static`.
```
error: lifetime may not live long enough
   --> tests/view/component-pass.rs:136:26
    |
107 |     create_scope_immediate(|cx| {
    |                             -- has type `BoundedScope<'1, '1>`
...
136 |         let _: View<G> = view! { cx, AttributesComponent(on:click = on_click) {} };
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cast requires that `'1` must outlive `'static`
    |
    = note: this error originates in the macro `$crate::view_with_elements` which comes from the expansion of the macro `view` (in Nightly builds, run with -Z macro-backtrace for more info)
```

We're doing that static transmute anyways in `event`, so I'll just put the same `SAFETY` above it to mark it.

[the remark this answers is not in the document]
wingertge: Actually I just recompiled and now the error is gone for some reason.

packages/sycamore/src/utils.rs
```diff
@@ -40,3 +46,135 @@ pub fn initial_node<G: GenericNode>(_el: &G) -> Option<View<G>> {
...
+            create_effect(cx, {
+                let name = name.clone();
+                move || el.set_attribute(name.clone(), Cow::Owned(s()))
+            });
+        }
+        AttributeValue::Bool(value) => {
+            let stringified = match value {
+                true => "true",
+                false => "false",
+            };
```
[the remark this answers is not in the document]
wingertge: If that's the case then your `view!` macro code is also wrong, it's copy pasted from that. Should I change it in both places or leave it for now?

packages/sycamore/src/utils.rs
```diff
@@ -40,3 +46,135 @@ pub fn initial_node<G: GenericNode>(_el: &G) -> Option<View<G>> {
...
+        }
+        AttributeValue::DangerouslySetInnerHtml(value) => {
+            el.dangerously_set_inner_html(value.into());
+        }
+        AttributeValue::DynamicDangerouslySetInnerHtml(value) => {
+            create_effect(cx, {
+                move || {
+                    el.dangerously_set_inner_html(Cow::Owned(value.to_string()));
+                }
+            });
```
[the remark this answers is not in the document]
wingertge: Yea I expected `rustfmt` to clean up that stuff, it was left over from when I was cloning something in the macro version. Thanks.

packages/sycamore/src/utils.rs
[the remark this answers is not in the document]
wingertge: Since it's a different situation (`view!` returns an entry in the static attributes, this util doesn't) I changed it in the utils. Might want to fix the `view!` macro implementation outside this PR.

wingertge:
I believe that's everything fixed now. No new changes aside from the suggested fixes and all tests are passing. This should be ready to merge now.

In case you didn't notice: I also added a real world use case example from my headless-ui port to the original PR text, in case there's any more questions about actual real world usage of this feature.

packages/sycamore/src/utils.rs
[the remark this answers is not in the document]
lukechu10: Let me take a look at that.

lukechu10 (approved):
Awesome! I think we can merge this now so that this doesn't become stale, and we can straighten out any issues we find later on.

Issue #506, opened 2022-10-16
Async event handlers


lukechu10:
To make it more async/await even more ergonomic in sycamore, it should be possible to use async directly inside event handlers. Currently, we need to use `spawn_local_scoped` inside the handler to use async. It would be nicer to directly allow event handlers to return a `impl Future<()>`.

Implementing this will probably take some type-level trickery since we can't naively implement a new `trait EventHandler` for both `Fn()` and `Fn() -> impl Future<()>`. The trait itself probably needs to be generic to by-pass this limitation.

# Desired result
```rust
view! { cx,
    button(on:click=|_| async { do_stuff().await; }) { "Click me!" }
}
```

PR #553, opened 2023-01-05
Typed event data + async event handlers

### LINKED ISSUE #241: Typed Event Handlers
Event handlers should not take a base `web_sys::Event` but rather the appropriate type for the type of event. See Yew docs from more information on what this should look like: https://yew.rs/concepts/html/events#event-types

lukechu10:
Closes #241 and #506 

~~Async event handlers somehow do not type check yet.~~

This is a breaking change because closures now accept different types as arguments depending on the event used. This is also a breaking change because invalid event names are no longer allowed, and the event structs must be used explicitly when using the builder API.

### DIFF
--- a/packages/sycamore/src/web/html.rs
+++ b/packages/sycamore/src/web/html.rs
@@ -417,6 +417,8 @@ macro_rules! define_events {
         )*
     ) => {
         $(
+            #[doc = concat!("The ", stringify!($ev), " event.")]
+            #[doc = stringify!($event)]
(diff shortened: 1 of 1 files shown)


lukechu10:
> @lukechu10 I am not sure if it's right place or time for my question :) ev seems so short for naming, maybe evt is better.

We can still change it but I am personally in favor of `ev`. It is quite common to see "ev" for an abbreviation of "event" in event handlers.

Issue #394, opened 2022-03-21
wasm bindgen example


g00nix:
A wasm bindgen example would be good for all cases where JavaScript is still required. One use case would be to interact with various browser extensions, like Metamask or Phantom App. This would look like this on the rust side:
```rust
#[wasm_bindgen(module = "/js/solana.js")]
extern "C" {
    #[wasm_bindgen(catch)]
    async fn connect_to_wallet() -> Result<JsValue, JsValue>;
}
```
And the javascript part in file `js/solana.js`:
```javascript
export async function connect_to_wallet() {
  try {
    const resp = await window.solana.connect();
    return resp.publicKey.toString();
  } catch (err) {
    return err;
  }
}
```

After this, the rust future can be called with `connect_to_wallet().await` and the `JsValue` has to be handled via sycamore.

lukechu10:
> I would like to do that.

Awesome! Also I think it would be better for the example if it did not pull in an extra JS library. Instead, it would probably be best if the JS were completely self-contained. That way, we keep the example nice and simple. Maybe just implement some kind of math function, or some DOM operation in JS that can be called in Rust.

If you have any questions, don't hesitate to ask me.

PR #558, opened 2023-01-17
Wasm bindgen example

wa1aric:
#394


examples/js-snippets/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+        span {(*product.get())}
+    }
+}
+
+fn main() {
+    sycamore::render(|cx| {
+        view! {
+            cx,
+            App
+        }
```
lukechu10, line 34: Nit:
```suggestion
        view! { cx,
            App
        }
```

examples/js-snippets/src/main.rs
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    view! {
+        cx,
+        input(type="number", bind:valueAsNumber=a)
+        span {"*"}
+        input(type="number", bind:valueAsNumber=b)
+        span {"="}
+        span {(*product.get())}
+    }
```
lukechu10, line 26: Nit: formatting
```suggestion
    view! { cx,
        input(type="number", bind:valueAsNumber=a)
        span { "*" }
        input(type="number", bind:valueAsNumber=b)
        span { "=" }
        span { (*product.get()) }
    }
```

lukechu10 (commented):
Thanks a lot!
Just two small formatting problems with the view! macro to make the style consistent with the rest of the codebase.

Also it might be better to name the JS file `index.js` instead of `functions.js` but that is a really small detail and probably doesn't really matter.

lukechu10:
Also just realized this. Could you please add an entry to `examples/README.md` in alphabetical order? Thanks!

lukechu10:
It seems like you haven't fully fixed the formatting. Notice the space after the `{` and before the `}`. Apart from that, everything looks good.

wa1aric:
Done!.

PR #559, opened 2023-01-17
Revert HTML templates and make everything typed using builder API

### LINKED ISSUE #556: Completely make elements type-checked. Type all the things!
This is the natural continuation of #354 and #553. We should strive to make Sycamore's elements system 100% type-checked for maximum type-safety and IDE auto-completion goodness.

The main thing that is not yet checked are HTML attributes. These should ideally also be completely typed to prevent typos and other bugs.

Also related to #554

lukechu10:
Closes #556 
Supersedes #537


lukechu10:
This PR has become a giant mess and so it will probably be easier to restart again, this time splitting it up into more incremental chunks. Also I really don't want to go through all these merge conflicts!

PR #560, opened 2023-01-21
adding `set_fn_mut` and `set_fn_mut_silent`

blainehansen:
More convenience functions :) These `set_fn_mut` variants use [`RefCell::borrow_mut`](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow_mut) and [`Rc::get_mut`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.get_mut) to directly mutably borrow the underlying value without a clone. 

I thought you would consider the possible panics acceptable since the `Modify` implementation does the same thing. It seems this is safe to do, since the borrow only lasts during the function's synchronous call and won't conflict with other borrows since `Rc` can't be sent across thread boundaries.

Am I crazy? There are tests covering it even with dependent signals, so it seems safe!

### DIFF
--- a/packages/sycamore-reactive/src/signal.rs
+++ b/packages/sycamore-reactive/src/signal.rs
@@ -247,6 +247,26 @@ impl<T> Signal<T> {
         self.set(f(&self.get_untracked()));
     }
 
+    /// Set the value of the state using a function that can mutate the value.
+    ///
+    /// This will notify and update any effects and memos that depend on this value.
+    ///
+    /// # Example
+    /// ```
+    /// # use sycamore_reactive::*;
+    /// # create_scope_immediate(|cx| {
+    /// let state = create_signal(cx, vec![]);
+    /// assert_eq!(*state.get(), vec![]);
+    ///
+    /// state.set_fn_mut(|&mut v| v.push(1));
+    /// assert_eq!(*state.get(), vec![1]);
+    /// # });
+    /// ```
+    pub fn set_fn_mut<F: FnMut(&mut T)>(&self, f: &mut F) {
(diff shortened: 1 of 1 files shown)


lukechu10:
> I thought you would consider the possible panics acceptable since the `Modify` implementation does the same thing. It seems this is safe to do, since the borrow only lasts during the function's synchronous call and won't conflict with other borrows since `Rc` can't be sent across thread boundaries.

Just for reference, the current `Modify` implementation can't panic because it creates a clone implicitly behind the hood. However, that being said, I would probably still consider this possible panic acceptable.

blainehansen:
This doesn't actually reliably work, I just tried to use it in a fairly simple case and it panicked. I'll close and open another pr if I can get this to work ha.

lukechu10:
> This doesn't actually reliably work, I just tried to use it in a fairly simple case and it panicked. I'll close and open another pr if I can get this to work ha.

I think to make this work, we would need to either get rid of the inner `Rc` in `RefCell<Rc>` or add another nested `RefCell` so `RefCell<Rc<RefCell>>`. I am personally leaning towards the former. I've been thinking maybe we can just make `.get()` only work for `Copy` types and add a `.read()` which wraps around `RefCell::borrow` and a similar `.write()`. We could also add a `.get_clone()` which would be an alias for `.read().as_ref().clone()`.

So to get the original behavior for `Signal` back, you'll need to add a `Rc::new(value)` to all the `create_signal`s and use `.get_clone()` instead of `.get()`.

Sadly, this would be a really massive breaking change again but I think it could be worth it.

PR #565, opened 2023-01-27
Add unique ID generation hook

wingertge:
This adds a way to generate IDs that are stable across the hydration boundary. If a hydration context is not available, it will fall back to a globally incremented ID.
This is useful for generating ids for accessible components and similar uses that require a stable globally unique ID.

# Example use case

```rust
let (component_id, id) = create_unique_id();
view! { cx,
    div(aria-labelledby = format!("{}_{id}", component_id.unwrap_or_default())) {
        // children
    }
}
```

The format boilerplate would likely be abstracted into a custom hook specific to the application, this just provides the primitive so the IDs are guaranteed to be globally unique.

### DIFF
--- a/packages/sycamore-core/src/lib.rs
+++ b/packages/sycamore-core/src/lib.rs
@@ -16,4 +16,5 @@ pub mod generic_node;
 pub mod hydrate;
 pub mod noderef;
 pub mod render;
+pub mod stable_id;
 pub mod view;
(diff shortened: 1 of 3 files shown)


wingertge:
The clippy errors aren't me, they're in files unrelated to this commit

lukechu10:
Thanks! I'm refactoring how the hydration system works in #559 but I'll merge this and I'll reimplement it in that other PR with the new hydration system.

PR #568, opened 2023-01-30
re-export `web_sys` for user convenience

blainehansen:
I found myself needing to annotate a handler event type, and didn't want to specify my own dependency on `web_sys`. This pull request should make patterns like the below easy:

```rust
use sycamore::web::web_sys;
let handle_enter = |event: web_sys::KeyboardEvent| {
	// ...
};
```

I also added `js_sys`. Are there any other libraries we should re-export? Is there any value in re-exporting `wasm_bindgen` for example?


arctic-hen7:
Maybe I'm missing something, but given `web-sys` is almost entirely controlled by feature flags, wouldn't this mean that users have to import it anyway almost always?

lukechu10:
> Maybe I'm missing something, but given `web-sys` is almost entirely controlled by feature flags, wouldn't this mean that users have to import it anyway almost always?

I agree with this sentiment about web-sys. It might make sense to re-export js-sys and wasm-bindgen, although I'm personally a bit iffy on this as well.

Yew used to have a web-sys re-export but they removed it: https://github.com/yewstack/yew/pull/2084
To be fair, they still continue to re-export the event types which they use and so maybe we should eb doing something similar?

Dav1dde:
I think sycamore should only re-export types that are also used in public signatures of sycamore's types and traits.

lukechu10:
Hi there, sorry for ignoring this for such a long time. I think what I would like to do is to have sycamore re-export `js-sys` and `wasm-bindgen` and have another module `events` which re-exports all the `web-sys` events only. Are you still interested in this? If so, please feel free to continue on this PR or open a new PR. If not, please let me know and I can take over. Thanks!

lukechu10:
Closed in favor of #642. @blainehansen I've included the original commits from this PR to give you credit.

PR #569, opened 2023-01-30
fixes bind macro problem

blainehansen:
When various `Signal` types are nested inside other structs, I wasn't able to use `bind:value` (the compiler would complain about `expected pattern` or something to that effect). I changed the proc macro codegen to fix it, and now this pattern works:

```rust
struct ThingRx {
	name: RcSignal<String>,
}
// ...

view!{cx,
	input(bind:value=c.name)
}
```


lukechu10:
Ah oops. Thanks for catching this.

lukechu10 (approved):

PR #570, opened 2023-02-03
Fix reconcile_fragments wrongly ordering nodes

alecmocatta:
I was seeing `reconcile_fragments` leave nodes in the wrong order. The `debug_assert_eq` at the bottom of my patch was failing.

This patch fixes the issue I was seeing.. but I'm still a bit suspicious of this code. Hopefully the debug assert will help surface any other bugs lurking here.

That latter `debug_assert_eq` assumes that `b` can be empty, is this is indeed the case?


lukechu10:
I think the extra assets make sense but I'm not sure the new logic does. In any case, it seems like some tests for `reconcile_fragments` are broken now. Also, do you have a test case for the original problem?

lukechu10:
Oh wait, I think the reason why the tests are failing is because of the new assert. It's because the fragment doesn't necessarily span the whole element. It's possible to have something like:
```rust
div {
    "before"
    (fragment)
    "after"
}
```
where the `"after"` shouldn't be part of the fragment.

alecmocatta:
Apologies, oversight on my part. Fixed and now passing CI, bar clippy on untouched code.

I'm afraid I didn't really grok the logic here so I don't have much confidence in my fix, nor the fn as a whole? I'd trust a simple fuzzer plus the assertions. Will add that if I have any more issues here!

My testcase was fragment `[a, b, c, d]` being replaced with `[a, e, c, f]`, which instead ended up with `[a, b, f, c]`.

lukechu10:
> with



> Apologies, oversight on my part. Fixed and now passing CI, bar clippy on untouched code.
> 
> I'm afraid I didn't really grok the logic here so I don't have much confidence in my fix, nor the fn as a whole? I'd trust a simple fuzzer plus the assertions. Will add that if I have any more issues here!
> 
> My testcase was fragment `[a, b, c, d]` being replaced with `[a, e, c, f]`, which instead ended up with `[a, e, f, c]`.

Thanks for the test case! I'll experiment a bit and let you know (hopefully soon).

lukechu10:
Sorry for taking so long to get back to you. It seems like your testcase already passes on master without your changes. Could you please verify this on your side? Once again, really sorry for not replying earlier.

lukechu10:
For reference, here is the test case which already passes on master:
```rust
#[wasm_bindgen_test]
fn reconcile_with_other_nodes_at_same_level() {
    let nodes = [
        DomNode::text_node("1".into()),
        DomNode::text_node("2".into()),
        DomNode::text_node("3".into()),
    ];
    let before = DomNode::text_node("before".into());
    let after = DomNode::text_node("after".into());
    let parent = DomNode::element::<html::div>();
    parent.append_child(&before);
    let child_nodes = nodes.to_vec();
    for node in &child_nodes {
        parent.append_child(node);
    }
    parent.append_child(&after);

    assert_text_content!(parent.to_web_sys(), "before123after");

    reconcile_fragments(&parent, &mut child_nodes.clone(), &child_nodes[..2]);
    assert_text_content!(parent.to_web_sys(), "before12after");
}
```
If this is not what you had in mind, please let me know.

alecmocatta:
Thanks for the response @lukechu10.

It's going to be a little while before we can update from [the version we're pinned to to `master`](https://github.com/alecmocatta/sycamore/compare/51622253cf29ac8e39cb982df8248627d43674d1...sycamore-rs:sycamore:master) – https://github.com/sycamore-rs/sycamore/pull/519 in particular will necessitate some refactoring.

However I'm quite sure this bug isn't fixed on `master` - the `debug_assert_eq`s in this PR were failing without the fix, and I don't see any other changes since that could have affected that. I'll try and work out a better testcase in the next couple days.

lukechu10:
Note that I've already merged the latest changes on master onto this branch. There aren't any merge conflicts right now.

alecmocatta:
Yes, sorry, to be clear we're pinned in our Cargo.lock to 51622253cf29ac8e39cb982df8248627d43674d1, and updating to the new head of this branch, or `master`, is nontrivial given https://github.com/sycamore-rs/sycamore/pull/519.

lukechu10:
Ah ok I understand now. Also note that you can replace all the places where `create_ref` is now a compile error with `create_ref_unsafe`, as that is essentially what `create_ref` used to be, although this is more of a temporary workaround rather than a proper fix.

lukechu10:
Closing because of inactivity. If this still persists after #626, please open a new issue.

Issue #572, opened 2023-02-08
Panic when using reactive value from context in scoped dynamic view


wingertge:
**Describe the bug**
When accessing a reactive value from a context inside a dynamic view in a child context, sycamore crashes from an effect loop.

**To Reproduce**

1. Create a child scope
2. Supply a context with reactive fields in that child scope
3. Use the child scope to render `Children` (`children.call(child_cx)`)
4. Access the reactive value from the context in the children
5. Update the reactive value
6. Crash

Since it's somewhat complicated, there is a minimal reproduction available here:
[dyn-view-bug/main.rs](https://github.com/wingertge/headlessui-sycamore/blob/master/examples/dyn-view-bug/src/main.rs)

**Expected behavior**
Update the view and don't crash

**Environment**

- Sycamore: master
- Browser: any
- OS: any

**Additional context**
The smallest diff between working and not working code is available here:
https://gist.github.com/wingertge/e47b90d57727f15a026d795ba5b7f8c5
This is the macro output generated by
```rust
view! { cx, div { (format!("{}", *context.0.get())) } }
```
(working)
and
```rust
view! { cx,
    div {
        (View::new_dyn(move || {
            format!("{}", *context.0.get()).to_view()
        }))
    }
}
```
(not working)

wingertge:
I managed  to create probably the smallest possible reproduction. Turns out the context has absolutely nothing to do with it, just accessing a signal in a dynamic view inside a child scope causes the crash

```rust
    let show = create_rc_signal(false);

    {
        let show = show.clone();
        spawn_local(async move {
            show.set(true);
        });
    }

    View::new_dyn_scoped(cx, move |cx| {
        let show = show.clone();
        View::new_dyn(cx, move || {
            show.track();
            View::empty()
        })
    })
```

PR #573, opened 2023-02-08
Fix Issue #572

wingertge:
This fixes the panic described in #572.

It removes the loop that unwraps inner dynamic views from the render code for `ViewType::Dyn`. I don't understand its purpose in the first place, no tests failed after removing it, TodoMVC still works exactly as expected and even my contrived example to specifically trigger that loop renders and updates perfectly fine. So I can only assume it was extraneous code left over from a refactor.


wingertge:
The clippy errors are still in other code, it's all the `format!` calls that could be update to the newer interpolation.

PR #575, opened 2023-02-11
Fix #562 / ignore the query parameters when matching a path to a route

### LINKED ISSUE #562: Navigate and a href fail to pass on query parameters
**Describe the bug**

In my application, I have a URL that takes query parameters. They are extracted by the router using JavaScript which works well. Now if I call my app with `/?filter=text` I expect it to receive the query parameter. That indeed works if called directly from the browser's URL bar (typing it manually). Coming from sycamore, however, this has two interesting variations:

* When calling `sycamore_router::navigate("/?filter=text")` I end up on my 404 page because apparently the router tries to resolve the full page by name instead of trying to resolve `/` and passing on all query parameters as is. The address bar of the browser, however, contains the correct address `/?filter=text`; manually reloading the page indeed gets me to the right page with all query parameters present.
* When using `a(href = "/?filter=text") { "Link" }` I end up on the page for `/` but with all query parameters cleared, i.e. the filter parameter did not make it to the routed endpoint.

Losing the query parameters while navigating or having to manually reload is a bug, in my opinion.

**To Reproduce**

```rust
pub fn get_query_parameter(key: &str) -> Option<String> {
    web_sys::UrlSearchParams::new_with_str(&web_sys::window()?.location().search().ok()?).ok()?.get(key)
}

pub enum Action {
    View(ViewParameters),
}

#[derive(Clone, Debug, Default)]
pub struct ViewParameters {
    pub filter: Option<String>,
}

impl TryFromSegments for Action {
    fn try_from_segments(param: &[&str]) -> Option<Self> {
        match param {
            [] => Some(Action::View(ViewParameters {
                filter: get_query_parameter("filter"),
            })),
            _ => None,
        }
    }
}
```

**Expected behavior**

I expect sycamore-router to be query-parameter-agnostic, i.e. even while #130 is unresolved I expect sycamore-router to pass on query parameters as is, neither ignoring them (the `a(href` case) nor leading to 404 pages (the `navigate` case).

Furthermore, navigate should not behave differently from what the `href` interceptor does on `a` tags in this regard.

**Environment**

- Sycamore: 0.8.2
- Browser: various chromium-based browsers
- OS: linux

Miroito:
fixes #562.

It's a simple problem with many possible solutions, I've decided to go with using the heap, it is possible to only use slices but the code becomes less readable.

Let me know if you've got a better idea.

### DIFF
--- a/packages/sycamore-router/src/lib.rs
+++ b/packages/sycamore-router/src/lib.rs
@@ -89,7 +89,13 @@ impl RoutePath {
     /// Attempt to match the path (url) with the current [`RoutePath`]. The path should already be
     /// split around `/` characters.
     pub fn match_path<'a>(&self, path: &[&'a str]) -> Option<Vec<Capture<'a>>> {
-        let mut paths = path.iter();
+        let mut paths = path.to_vec();
+        if let Some(last) = paths.last_mut() {
+            if let Some(pos) = last.find('?') {
+                *last = &last[..pos];
+            }
+        }
+        let mut paths = paths.iter();
         let mut segments = self.segments.iter();
         let mut captures = Vec::new();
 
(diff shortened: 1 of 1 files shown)


arctic-hen7:
@lukechu10  any chance of this being merged soon? Also, does this PR fix the same issue for hash parameters?

lukechu10:
Sorry for ignoring this PR for so long! No I don't think this fixes hash parameters but I'll merge this first now.

lukechu10 (approved):

PR #589, opened 2023-03-19
Centered badges

simonhyll:
Call me crazy but I think this looks better


lukechu10:
You're absolutely right! Thanks!

PR #596, opened 2023-04-03
Enable codecov region coverage

lukechu10:


lukechu10:
Will be closing until Codecov's region coverage support improves.

PR #599, opened 2023-04-13
Reduce allocations and dynamic dispatch in effects and memos

lyphyser:
This series of commits removes unnecessary allocations in effects and memos and makes calling effect callbacks only do a single dynamic dispatch rather than two.


lukechu10:
Supserseded by #626

PR #647, opened 2023-11-26
Update Router example to use non-referenced ReadSignal

noxxxxxious:
Also rerendered docs


lukechu10 (changes requested):
There's another instance of the same problem on Line 251.

Also could you remove the `.as_ref()` while you're at it? Thanks!

noxxxxxious:
That should do it!

lukechu10 (approved):
Awesome thanks!

Issue #648, opened 2023-12-02
Book (website) routing not properly updating


lukechu10:
**Describe the bug**
When navigating to a new page in the Book, there is sometimes what seems to be a race-condition which causes the following visual bug to happen:
![image](https://github.com/sycamore-rs/sycamore/assets/37006668/84b5e527-f6c5-46d4-aa46-7273551b9a6b)
[Relevant Discord discussion](https://discord.com/channels/820400041332179004/820400041332179009/1173039854007877682)

**To Reproduce**

1. Go to https://sycamore-rs.netlify.app/docs/getting_started/hello_world
2. Click on another page from the sidebar, e.g. "Hello World!"
3. (Maybe) observe the visual bug.

**Environment**

- Sycamore: 0.9.0-beta.2, current master (16acf8e)
- Browser: Edge 119

Saplyn:
It seems that the book works normal now. *Edge Version 121.0.2277.4 (Official build) stable app, beta channel (64-bit)*

lukechu10:
The annoying part is that I'm only able to reproduce this sometimes and always when I think I've solved the issue. So I'll leave this open for now until I'm more certain that it's fixed.

steven89:
I can consistently reproduce on `Chromium: 122.0.6261.69`:
1. navigate to [https://sycamore-rs.netlify.app/](https://sycamore-rs.netlify.app/) (navigating directly to the home page or reloading on the homepage if I come from another page)
2. press either
  - "Get Started" button in the middle of the page
  - "Book" menu at the top
3. Navigating within the docs with the links now creates rendering issue, as well as scrolling the left-side menu
4. reloading the page while in the docs fixes the rendering

Left side menu scrolling issue:
<img width="193" alt="Screenshot 2024-02-28 at 10 02 55 AM" src="https://github.com/sycamore-rs/sycamore/assets/609271/b4fca317-a9b8-47e7-9a6e-85afb5bffb32">

steven89:
Also once in the docs (after a reload) clicking on a `latest version` link ("For up-to-date documentation, see the [latest version](https://sycamore-rs.netlify.app/docs/v0.8/getting_started/installation) (v0.8).") restarts the rendering issue, again reloading the page afterwards fixes it.

Hmikihiro:
I can reproduce on Firefox 123.0 for MacOS.

When I clicked 'Get Started', the `<div class="flex w-full">` tag was created twice.
[...]

PR #661, opened 2024-03-11
Fix creating <div> twice on website

Hmikihiro:
#648

I might fix the bug.
`view!{}` was drawn two times when calling the code `if let Some(cached_sidebar_data) = cached_sidebar_data.get_clone(){view!{...}}`.
I think that the bug is around `cached_sidebar_data`.
For the reason, instead of calling from `fetch_docs_data`, I changed to `cached_sidebar_data` is initialized only once.

### DIFF
--- a/website/src/main.rs
+++ b/website/src/main.rs
@@ -89,16 +89,7 @@ fn switch<G: Html>(route: ReadSignal<Routes>) -> View<G> {
             view! {
                 (if let Some(data) = data.get_clone() {
                     if let Some(cached_sidebar_data) = cached_sidebar_data.get_clone() {
-                        view! {
-                            content::Content(
-                                data=data.clone(),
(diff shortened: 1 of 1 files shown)


Hmikihiro:
Sorry, I made a mistake whit applying the formatting.

lukechu10 (commented):
Sorry for taking so long to respond... I'm afraid that while this solution works for the latest docs, it does not work for historical versions of the docs which can be accessed by pressing on the version label in the top-left corner. The original idea was to refresh the sidebar data every time the page changed so that we will always get the latest up to date version of the list of all the pages.

But this is already much better than what is happening right now so I'm inclined to just merge this as is and fix this in another PR. The website definitely needs a major overhaul but that can probably be left for another PR.

PR #674, opened 2024-06-03
Replace CountAPI as it got taken down. 

JasonLovesDoggo:
Replaced CountAPI as it got bought out then taken down by APILayer, this is a fully compliant replacement for it. 

This code was created in https://github.com/sycamore-rs/sycamore/pull/305


lukechu10:
Awesome thanks!

Issue #676, opened 2024-07-10
rust-analyzer reports a warning arising from proc-macro


mtshr:
**Describe the bug**

rust-analyzer reports a warning arising from the proc-macro.

**To Reproduce**

Follow the hello world example and write the source code in Visual Studio Code with rust-analyzer.
https://sycamore-rs.netlify.app/docs/getting_started/hello_world

**Expected behavior**

No warnings emitted.

**Screenshots**

![Screenshot from 2024-07-10 21-16-27](https://github.com/sycamore-rs/sycamore/assets/10617402/60ee06ce-df42-408b-bec3-62f64ec77d3d)

**Environment**

- Sycamore: "0.9.0-beta.2"
- Browser: N/A
- OS: Ubuntu 22.04.4 LTS
- rust-analyzer: v0.3.2029
- Visual Studio Code: 1.91.0

**Additional context**

This warning is not emitted during the build process with cargo, so is not caught in github actions.

**Suggestion**

Basically, I do not want to put an `[allow(non_snake_case_types)]` attribute in my source code, so let me suggest adding the attribute to the type alias in question in `sycamore-macro` crate, or renaming the type alias like below.

https://github.com/sycamore-rs/sycamore/blob/90356a7851d5222eecbd04a2a5b606ee4663d1a7/packages/sycamore-macro/src/view.rs#L157-L166

```diff
                quote! {{
-                    type __tag = ::sycamore::web::html::#tag;
+                    type __Tag = ::sycamore::web::html::#tag;
                    ::sycamore::generic_node::TemplateShape::Element {
-                        tag: __tag::TAG_NAME,
-                        ns: __tag::NAMESPACE,
+                        tag: __Tag::TAG_NAME,
+                        ns: __Tag::NAMESPACE,
                        children: &[#(#children),*],
                        attributes: &[#(#attrs),*],
                        flag: #flag,
                    }
                }}
```

PR #677, opened 2024-07-10
Rename the tag type alias

mtshr:
The suggested change in #676 

The change would suppress the unwanted warning from rust-analyzer.

Note: clippy seems to have been failing for at least 3 weeks.
https://github.com/sycamore-rs/sycamore/actions/workflows/test.yml


lukechu10:
Thanks!

PR #705, opened 2024-09-27
Move `MaybeDyn` to `sycamore-reactive`

lukechu10:
Also slightly changes how we create `MaybeDyn` values to work around specialization.


lukechu10:
Unfortunately, this change requires adding a (optional) dependency on `wasm-bindgen` in `sycamore-reactive` because of Rust's orphan rules, which somewhat dilutes the framework-agnostic architecture I was trying to go with. However, I don't think this is a deal breaker so I'll keep this solutino for now.

Issue #703, opened 2024-09-25
Allow arbitrary attribute names by using string literal


lukechu10:
Allow arbitrary attribute names with the following syntax:
```rust
view! {
    div("myattribute"="abc") { ... }
}
```

This used to work since we did not type check attributes but no longer works since #679 

This should codegen to the `.attr(name, value)` method on the builder API.

PR #707, opened 2024-09-30
Allow attribute names by using string literal

Kromgart:
Added a new PropType variant and the corresponding updates to parsing and dom manipulation. 

Added 1 case into 'element-pass' test using such quoted-string attribute.

Updated docs to show usage of "quoted string-literal" attribute name.

Note, that the issue #703 says _"Allow arbitrary attribute names by using string literal."_, but we are using [web_sys::Element::set_attribute](https://github.com/sycamore-rs/sycamore/blob/b6577326590257fef8268184d547ecce840b3d2a/packages/sycamore-web/src/node/dom_node.rs#L119), and that call will panic if you try to set attribute name to something illegal. It seems to get upset by anything except alphanumeric characters, underscores or non-leading-dashes.

But at least now you can add such horrible, but valid attributes, if needed:

```rust
button("hello-22--"="no-way") {}
```

### DIFF
--- a/docs/next/basics/view.md
+++ b/docs/next/basics/view.md
@@ -93,7 +93,7 @@ Attributes (including classes and ids) can also be specified.
 
 ```rust
 view! {
-    p(class="my-class", id="my-paragraph", aria-label="My paragraph")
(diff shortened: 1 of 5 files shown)


lukechu10 (approved):
LGTM! Thanks!

PR #711, opened 2024-10-01
Fix #704 - Updating the check whether a view must be dynamic.

### LINKED ISSUE #704: Better heuristics for dynamic expressions in view macro
Right now, we have a very simple heuristic for deciding whether an expression used in the `view!` macro is dynamic or not. That heuristic is:

- Expression is a single identifier, e.g. `my_signal`.
- Expression is a literal, e.g. `123` or `"abc"`.
- Expression is a closure, e.g. `move || "abc"`. This is also to avoid creating a closure that returns a closure.

Anything that is not one of those is automatically wrapped inside a closure so that it becomes dynamic. We should improve this heuristic to also allow expressions of the form:

- Expression is a field/tuple access expression, e.g. `my.prop.value`. (Note that `.` operator is left associative so we can write this recursively as well).
- Tuple expressions, e.g. `(a, b, c)`
- Struct expressions, e.g. `MyStruct { a, b, c }`.
- Array expressions, e.g. `[a, b, c]`.

where `a`, `b`, and `c` are non dynamic. And possible even more kinds of expressions (see Rust reference).

Finally, when this is fixed, these should be updated:
https://github.com/sycamore-rs/sycamore/blob/9316bc38429789849a39513ec37b705e5dc55744/examples/attributes-passthrough/src/main.rs#L19

https://github.com/sycamore-rs/sycamore/blob/9316bc38429789849a39513ec37b705e5dc55744/examples/js-framework-benchmark/src/main.rs#L167

as well as possibly a few others.

Relevant code is in this file:
https://github.com/sycamore-rs/sycamore/blob/main/packages/sycamore-macro/src/view.rs

Kromgart:
Fix for #704.

The following expressions will be considered as non-dynamic (NDY):

- literal
- closure
- path
- field accessor
```rust
document.author.id
```
- parentheses, where inner expression is NDY
```rust
(expr)
```
- tuple, where elements are NDY
```rust
(a, b, c)
```
- array, where elements or length are NDY
```rust
[a, b, c]
[d; len]
```
- struct, where field values are NDY
```rust
Doc {
    id: a,
    author: Author {
        id: (b, c),
    },
}
```
- type cast, where expression is NDY
```rust
expr as sometype
```
- **view!** macro
- block, when either a const block, or all statements are NDY
- **loop**, **while** and **for**, (when all their parts are NDY)
- **break**, **continue**
- **let**, when pattern and expression are NDY
- **match**, when the matched expression, and all arms (patterns, guards and bodies) are NDY
- **if**, when the condition and branches are NDY
- unary,binary, index and range expressions, when composed from NDY

Everything else is still considered dynamic.

### DIFF
--- a/examples/attributes-passthrough/src/main.rs
+++ b/examples/attributes-passthrough/src/main.rs
@@ -18,7 +18,6 @@ fn CustomButton(props: CustomButtonProps) -> View {
 
     let children = props.children.call();
     view! {
-        // TODO: Remove the .clone() here.
         button(id=props.id, ..props.attributes) {
             (children)
         }
--- a/examples/attributes-passthrough/src/main.rs
+++ b/examples/attributes-passthrough/src/main.rs
@@ -19,7 +19,7 @@ fn CustomButton(props: CustomButtonProps) -> View {
     let children = props.children.call();
     view! {
         // TODO: Remove the .clone() here.
-        button(id=props.id.clone(), ..props.attributes) {
+        button(id=props.id, ..props.attributes) {
             (children)
         }
     }
--- a/packages/sycamore-macro/src/view.rs
+++ b/packages/sycamore-macro/src/view.rs
@@ -198,61 +198,26 @@ fn is_component(ident: &TagIdent) -> bool {
 }
 
 fn is_dyn(ex: &Expr) -> bool {
-    fn is_dyn_macro(m: &syn::Macro) -> bool {
-        // Bodies of nested inner view! macros will be checked for dynamic
-        // parts when their own codegen is run.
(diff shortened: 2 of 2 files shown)


Kromgart:
I updated the attributes-passthrough example, as mentioned in the issue. But I'm not sure what should be done to the second mentioned example (js-framework-benchmark).

lukechu10:
Ah sorry I don’t think anything can be done about the js-framework-benchmark example yet. From a quick look, this looks great. I’ll try to take a closer look soon.

lukechu10 (approved):
LGTM!

Issue #447, opened 2022-06-30
Add support for optional dyn attributes in the `view!` macro


mc1098:
Supporting optional dyn attributes would bring the `view!` macro inline with the builder API:
https://github.com/sycamore-rs/sycamore/blob/b6414fa9ffe6debc9245e0df5111e89eeece2fea/packages/sycamore/src/builder.rs#L162-L178

This allows for not only dynamic changes to the attribute value, which is currently supported, but also allows for the attribute to be removed when the optional value is `None`.

___

## Unresolved questions

 1. Should dynamic attributes in the `view!` macro accept the same arg as the builder: `FnMut() -> Option<&str> + 'a`?
 This would save some of the trait kung fu and the implementation of each type specifically to avoid conflicting impls.
 2. When the value is for the `class` attribute the `view!` macro sets the class name so if this is `None` should it set the class name to any empty string?
 `view!` macro uses `set_class_name`:
 https://github.com/sycamore-rs/sycamore/blob/b6414fa9ffe6debc9245e0df5111e89eeece2fea/packages/sycamore-macro/src/view/codegen.rs#L284-L288
 yet the builder API uses the add and remove class with a boolean yielding closure to decide:
 https://github.com/sycamore-rs/sycamore/blob/b6414fa9ffe6debc9245e0df5111e89eeece2fea/packages/sycamore/src/builder.rs#L292-L307

PR #712, opened 2024-10-07
Allow optional attribute values

### LINKED ISSUE #622: Optional attributes in view!
It would be great if the `view!` macro allowed attributes to be set to an `Optional`, and then would generate the attribute if and only if the Optional is Some.

**Use case:**
Let's say we are wrapping a `input` component into our own `Switch` component:
```rust

#[derive(Prop)]
pub struct SwitchProps<'cx> {
    checked: &'cx Signal<bool>,
    // id can be present or absent
    #[builder(default)]
    id: Option<&'cx str>,
}

#[component]
pub fn Switch<'a, G: Html>(cx: Scope<'a>, props: SwitchProps<'a>) -> View<G> {
    view! {cx,
        input(
            type="checkbox",
            role="switch",
            class="...",
            id=(props.id),
            bind:checked=props.checked,
        )
    }
}
```
The `SwitchProps::id` prop will be present or absent depending on whether the call site specified it. But the `view!` macro currently does not allow it being passed through to `input` the same way. Instead, we have to write much more verbosely:
```rust
#[component]
pub fn Switch<'a, G: Html>(cx: Scope<'a>, props: SwitchProps<'a>) -> View<G> {
    if let Some(id) = props.id {
        view! {cx,
            input(
                type="checkbox",
                role="switch",
                class="...",
                id=id,
                bind:checked=props.checked,
            )
        }
    } else {
        view! {cx,
            input(
                type="checkbox",
                role="switch",
                class="...",
                bind:checked=props.checked,
            )
        }
    }
}
```

Kromgart:
- Added .attr_opt() method into builder-like API.

- Allow optional attributes in view! macro, the syntax is: 
  `<attr-name>?=<optional_value>`

- If the attribute value is None, the attribute is removed, just like a bool attribute does with a `false` value.

This should fix #622 (and maybe #447?)

Example:

```rust
fn main() {
    sycamore::render(|| {
        type SigOptStr<'a> = ReadSignal<Option<Cow<'a, str>>>;

        let the_class = Some("extra-dark");
        let sig_counter: Signal<i32> = create_signal(0);

        let on_btn_click = move |_| {
            sig_counter.set(sig_counter.get() + 1);
        };

        let sig_style: SigOptStr = sig_counter.map(|i| {
            if i % 2 == 1 {
                Some("color: red;".into())
            } else {
                None
            }
        });

        let sig_main_id: SigOptStr = sig_counter.map(|i| {
            if i % 2 == 0 {
                Some(i.to_string().into())
            } else {
                None
            }
        });

        let sig_data_id: SigOptStr = sig_counter.map(|i| {
            if i % 3 == 0 {
                Some(i.to_string().into())
            } else {
                None
            }
        });

        view! {
            button(on:click=on_btn_click) { "Toggle" }
            p(
                // Quoted attribute, dynamic non-option value
                "counter"=sig_counter.to_string(),
                // Plain attribute, non-dynamic option value
                class?=the_class,
                // Plain attribute, dynamic option value
                id?=sig_main_id,
                // Dashed attribute, dynamic option value
                data-id?=sig_data_id,
                // Quoted attribute, dynamic option value
                "style"?=sig_style,
            ) {
                "Some text"
            }
        }
    });
}
```

If this approach is acceptable - let me know, I'll add some tests into sycamore-macro (but if some other tests would be needed - let me know where).

### DIFF
--- a/packages/sycamore-macro/src/view.rs
+++ b/packages/sycamore-macro/src/view.rs
@@ -99,14 +99,21 @@ impl Codegen {
             quote! { #value }
         };
         match &attr.ty {
-            PropType::Plain { ident } => {
-                quote! { .#ident(#dyn_value) }
+            PropType::Plain { ident, optional } => {
+                if *optional {
+                    let ident = ident.to_string();
+                    quote! { .attr_opt(#ident, #dyn_value) }
+                } else {
+                    quote! { .#ident(#dyn_value) }
+                }
             }
-            PropType::PlainHyphenated { ident } => {
-                quote! { .attr(#ident, #dyn_value) }
-            }
-            PropType::PlainQuoted { ident } => {
-                quote! { .attr(#ident, #dyn_value) }
+            PropType::PlainHyphenated { ident, optional }
+            | PropType::PlainQuoted { ident, optional } => {
+                if *optional {
+                    quote! { .attr_opt(#ident, #dyn_value) }
+                } else {
+                    quote! { .attr(#ident, #dyn_value) }
+                }
             }
             PropType::Directive { dir, ident } => match dir.to_string().as_str() {
                 "on" => quote! { .on(::sycamore::rt::events::#ident, #value) },
@@ -139,7 +146,7 @@ impl Codegen {
         let plain = props
             .iter()
             .filter_map(|prop| match &prop.ty {
-                PropType::Plain { ident } => Some((ident, prop.value.clone())),
+                PropType::Plain { ident, optional: _ } => Some((ident, prop.value.clone())),
                 _ => None,
             })
             .collect::<Vec<_>>();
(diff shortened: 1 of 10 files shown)


lukechu10:
I was actually thinking of a bit of a different design where we can just keep the existing syntax but allow passing `Option<T>` as well to make it optional.

This would be done by changing the type of string attributes to something like `Option<Cow<'static, str>>` and automatically wrapping string values in `Some(...)`.

We already have the machinery to do this with `MaybeDyn`.

Kromgart:
Ah, ok then, no problem! :)

> I was actually thinking of a bit of a different design where we can just keep the existing syntax but allow passing `Option<T>` as well to make it optional.

Yeah, that would look better and that was what I've tried initially, but got bogged down in untangling conflicting impls and tried this approach.

PR #720, opened 2024-10-15
fix(macro): handle patterns in inline_props component arguments

### LINKED ISSUE #718: Inline props does not properly support patterns
**Describe the bug**
Using `#[component(inline_props)]` does not properly account for using patterns in function parameters.

**To Reproduce**
```rust
#[component(inline_props)]
fn Foo(mut value: i32) -> View {
    view! {}
}
```

The problem occurs when generating code for the prop struct. This is what ends up being generated currently:
```rust
#[derive(Props)]
struct FooProps {
    mut value: i32
}
```
where the `mut` is obviously out of place.

**Expected behavior**
Patterns work seamlessly with `#[component(inline_props)]`

**Environment**

- Sycamore: master

KekmaTime:
closes #718 

The fix removes  pattern modifiers when generating the props struct, ensuring valid syntax 
while preserving intended mutability semantics within the component function.

### DIFF
--- a/packages/sycamore-macro/src/component.rs
+++ b/packages/sycamore-macro/src/component.rs
@@ -261,11 +261,31 @@ fn inline_props_impl(item: &mut ItemFn) -> Result<TokenStream> {
     let props_struct_ident = format_ident!("{}_Props", item.sig.ident);
 
     let inputs = item.sig.inputs.clone();
-    let props = inputs.into_iter().collect::<Vec<_>>();
+    let props = inputs
+        .into_iter()
+        .map(|arg| match arg {
+            FnArg::Typed(pat_type) => {
+                // Remove any mut or other pattern modifiers
+                let pat = match *pat_type.pat {
+                    Pat::Ident(mut ident_pat) => {
+                        ident_pat.mutability = None;
+                        Pat::Ident(ident_pat)
+                    }
+                    // Handle other patterns if necessary
+                    _ => panic!("Unsupported pattern type in inline_props"),
+                };
+                syn::PatType {
+                    pat: Box::new(pat),
+                    ..pat_type
+                }
+            }
+            FnArg::Receiver(_) => panic!("Components cannot have receivers"),
+        })
+        .collect::<Vec<_>>();
 
     let generics = &item.sig.generics;
     let generics_phantoms = generics.params.iter().enumerate().filter_map(|(i, param)| {
-        let phantom_ident = format_ident!("__phantom{i}");
+        let phantom_ident = format_ident!("__phantom{}", i);
         match param {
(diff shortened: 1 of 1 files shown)


packages/sycamore-macro/src/component.rs
```diff
@@ -261,11 +261,31 @@ fn inline_props_impl(item: &mut ItemFn) -> Result<TokenStream> {
...
+                }
+            }
+            FnArg::Receiver(_) => panic!("Components cannot have receivers"),
+        })
+        .collect::<Vec<_>>();
 
     let generics = &item.sig.generics;
     let generics_phantoms = generics.params.iter().enumerate().filter_map(|(i, param)| {
-        let phantom_ident = format_ident!("__phantom{i}");
+        let phantom_ident = format_ident!("__phantom{}", i);
```
lukechu10, line 288: I'm not sure why you changed this

lukechu10 (commented):
Ah I've just realized that supporting arbitrary patterns for inline props means that we no longer know what the prop name is.

Maybe we can only support `mut/ref` patterns and `{xyz} @ ...` patterns so that we always have an identifier we can use (where `{xyz}` is any binding, so we need to get rid of `mut/ref` again).

lukechu10:
Also it would be great if you could add some tests to `packages/sycamore-macro/tests/component/inline-props-pass.rs`

packages/sycamore-macro/src/component.rs
KekmaTime: > I'm not sure why you changed this

ok that was a mistake on my part ill fix it ASAP

KekmaTime: > I'm not sure why you changed this

i think that is the wrong way of formatting. `format_ident!("__phantom{i}")` treats` {i}` as a literal, so the identifier becomes `__phantom{i}` exactly, without replacing i with its value.

KekmaTime:
> Also it would be great if you could add some tests to `packages/sycamore-macro/tests/component/inline-props-pass.rs`

sure ill get to it right away 👍🏽

packages/sycamore-macro/src/component.rs
lukechu10: I'm pretty sure the `{i}` works because it generates the right code. Also `format_ident!` uses `std::format!` under the hood and the new `format!()` syntax is supported since Rust 1.58.

KekmaTime: > I'm pretty sure the `{i}` works because it generates the right code. Also `format_ident!` uses `std::format!` under the hood and the new `format!()` syntax is supported since Rust 1.58.

im sorry i didnt know that 😓

PR #721, opened 2024-10-15
Disable website publishing workflow

lukechu10:
This is to prepare for moving to the new docsite which will be built in a new repo.


lukechu10:
Closed in favor of #726

Issue #507, opened 2022-10-16
Support `impl Trait` syntax with `inline_props`


lukechu10:
Right now, `inline_props` does not support the `impl Trait` syntax which is often useful to prevent adding tons of generics to a function.

We would like the following syntax to work out of the box:
```rust
#[component(inline_props]
fn MyComponent<G: Html>(cx: Scope, foo: impl MyTrait) -> View<G> {
    ...
} 
```

The difficulty with this is that we cannot directly translate `impl MyTrait` into a struct field because this syntax is only supported for functions. We therefore need to manually transform the `impl MyTrait` into an explicit generic parameter in the generated struct.

davidon-top:
I've done something similar in [rstml-component](https://github.com/YoloDev/rstml-component/pull/15), i will make a pr once i have a starting point.

PR #746, opened 2024-11-01
Support `impl Trait` syntax with `inline_props`

davidon-top:
closes: #507

### DIFF
--- a/packages/sycamore-macro/src/inline_props.rs
+++ b/packages/sycamore-macro/src/inline_props.rs
@@ -1,7 +1,6 @@
-use proc_macro2::{Span, TokenStream};
+use proc_macro2::Span;
 use quote::format_ident;
 use syn::punctuated::Punctuated;
-use syn::token::Comma;
 use syn::{
     Field, GenericParam, Generics, Ident, Path, PathArguments, PathSegment, Token, Type,
     TypeImplTrait, TypeParam, TypePath, Visibility,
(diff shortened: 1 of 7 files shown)


packages/sycamore-macro/src/inline_props.rs
```diff
@@ -0,0 +1,108 @@
...
+pub fn resolve_type(generics: &mut Generics, ty: Type) -> Type {
+    match ty {
+        Type::ImplTrait(inner) => add_generic(generics, inner),
+        Type::Array(inner) => {
+            let elem = resolve_type(generics, *inner.elem);
+            Type::Array(syn::TypeArray {
+                bracket_token: inner.bracket_token,
+                elem: Box::new(elem),
+                semi_token: inner.semi_token,
+                len: inner.len,
```
lukechu10, line 22: I think you could probably use the spread syntax to spread `inner` instead of duplicating all the fields. That would probably be a bit cleaner than this.

Same comment goes for other match branches

lukechu10 (commented):
Thanks this looks great! Just a small comment

davidon-top:
I don't know how the code coverage works for macros but i did write tests for the untested parts according to the bot but it didn't help much with the %

lukechu10:
> I don't know how the code coverage works for macros but i did write tests for the untested parts according to the bot but it didn't help much with the %

I wouldn't worry too much about the code coverage. It's pretty noisy and doesn't worry with the wasm test suite anyways.

lukechu10 (approved):
LGTM!

PR #749, opened 2024-11-01
Router example

davidon-top:
I was going to tackle the remaining 0.9 issues, half of which are router related, and i noticed that there is no example using router (which would be useful for router development).


lukechu10 (approved):
LGTM Thanks!

lukechu10:
Wait actually could you add an entry to `examples/README.md`

davidon-top:
good call, forgot about that

lukechu10 (approved):
Awesome!

Issue #699, opened 2024-09-19
Allow adding derives to generated prop type for `inline_props`


lukechu10:
Sometimes, the generated prop type needs to implement additional traits which can be derived using a derive macro. This arises in, e.g. [mdsycx](https://github.com/lukechu10/mdsycx) where the props type needs to derive `FromMd` so that the component can be used from Markdown.

Currently, we cannot use `inline_props` in this case because we have no way of specifying addition derives on the generated prop type.

# Proposal

Allow `#[component(inline_props, derive(FromMd))]` which adds a `#[derive(FromMd)]` onto the generated type.

Possibly even allow arbitrary attributes to be passed in in the same position which would forward them onto the generated type.

davidon-top:
will make a pr

PR #750, opened 2024-11-01
Allow adding derives to generated prop type for `inline_props`

davidon-top:
closes: #699

### DIFF
--- a/packages/sycamore-macro/src/component.rs
+++ b/packages/sycamore-macro/src/component.rs
@@ -275,17 +275,21 @@ pub struct ComponentArgs {
 impl Parse for ComponentArgs {
     fn parse(input: ParseStream) -> Result<Self> {
         let inline_props: Option<Ident> = input.parse()?;
-        if let Some(inline_props) = &inline_props {
+        let (comma, attrs) = if let Some(inline_props) = &inline_props {
             // Check if the ident is correct.
             if *inline_props != "inline_props" {
                 return Err(Error::new(inline_props.span(), "expected `inline_props`"));
             }
-        }
-        let comma: Option<Token![,]> = input.parse()?;
-        let attrs: Punctuated<ComponentAttrArgs, Token![,]> = if comma.is_some() {
-            input.parse_terminated(ComponentAttrArgs::parse, Token![,])?
+
+            let comma: Option<Token![,]> = input.parse()?;
+            let attrs: Punctuated<ComponentAttrArgs, Token![,]> = if comma.is_some() {
+                input.parse_terminated(ComponentAttrArgs::parse, Token![,])?
+            } else {
+                Punctuated::new()
+            };
+            (comma, attrs)
         } else {
-            Punctuated::new()
+            (None, Punctuated::new())
         };
         Ok(Self {
             inline_props,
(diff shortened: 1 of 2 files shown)


packages/sycamore-macro/src/component.rs
```diff
@@ -307,10 +368,12 @@ fn inline_props_impl(item: &mut ItemFn) -> Result<TokenStream> {
 
     let doc_comment = format!("Props for [`{}`].", item.sig.ident);
 
+    let attrs = attrs.into_iter().collect::<Vec<_>>();
```
lukechu10, line 371: I don’t think you need this line since quote! can take iterators directly already.

packages/sycamore-macro/src/component.rs
```diff
@@ -233,14 +281,24 @@ impl Parse for ComponentArgs {
                 return Err(Error::new(inline_props.span(), "expected `inline_props`"));
             }
         }
-        Ok(Self { inline_props })
+        let comma: Option<Token![,]> = input.parse()?;
```
lukechu10, line 284: I think we only want to parse this if we have inline props.

packages/sycamore-macro/tests/component/inline-props-pass.rs
```diff
@@ -75,4 +75,9 @@ fn PropsWithVariousImplGenerics(
     }
 }
 
+#[component(inline_props, derive(Clone))]
```
lukechu10, line 78: Could you add a dummy prop field and also test that the prop actually implements the Clone trait? You can create an instance of the prop type with the automatically generated builder API.

lukechu10 (changes requested):
Thanks! I made a few comments here and there. I’m currently on my phone so afraid I can’t check what I said.

packages/sycamore-macro/src/component.rs
davidon-top: it has to be there because before this line attrs is Punctuated<_, Token![,]> so the macro would otherwise produce #[derive(Clone)]**,**#[derive(Debug)] with a , in between which would result in an error

packages/sycamore-macro/src/component.rs
```diff
@@ -219,28 +221,88 @@ impl ToTokens for ComponentFn {
...
+        let paren = parenthesized!(content in input);
+        let tokens = content.parse()?;
+        Ok(Self { paren, tokens })
+    }
+}
+
+pub struct ComponentAttrArgs {
+    ident: Ident,
+    call: ParenthesizedTokens,
+}
```
lukechu10, line 248: Now that I look at this, I think this struct should be replaced with just [`syn::Meta`](https://docs.rs/syn/latest/syn/enum.Meta.html) which is also more general since it also allows `#[path]` and `#[attr = value]`

packages/sycamore-macro/src/component.rs
```diff
@@ -307,10 +372,12 @@ fn inline_props_impl(item: &mut ItemFn) -> Result<TokenStream> {
 
     let doc_comment = format!("Props for [`{}`].", item.sig.ident);
 
+    let attrs = attrs.into_iter().collect::<Vec<_>>();
```
lukechu10, line 375: We could then change this to
```rust
attrs.into_iter().map(|attr| Attribute {
    pound_token: parse_quote!(#),
    // ...
}).collect::<Vec<_>>()
```
so that we generate the `#[...]` brackets around the attribute. (see previous comment)

packages/sycamore-macro/src/component.rs
davidon-top: Oh ye, i only thought of valid rust macros but other macros applied to the struct means that those macros may require other attribute styles

lukechu10 (approved):
Looks great!

Issue #335, opened 2022-01-23
Router reload


arctic-hen7:
Right now, Sycamore's router supports navigating to  a new page and also replacing the current entry in the history. Extending on this, it would be useful if the router could reload the current route. At first, I know this seems completely pointless, but, as I've discovered through Perseus' [state freezing](https://github.com/arctic-hen7/perseus/issues/120), it's not! Right now, if you call `navigate(url)` where `url` is the same as the current page, the code in the `Router` callback doesn't run, nor should it.

However, there are cases, particularly for high-level contexts that have been updated, in which reloading a page with new state is extremely useful. This seems to be the offending code:

https://github.com/sycamore-rs/sycamore/blob/86343315e8cd921f8c4618b6dc32fc0191ddceba/packages/sycamore-router/src/router.rs#L112-L113

While I accept that this is a great default, it would be very useful to have a function that bypasses it to enable reloading the current route. Notably, this would probably require a not insignificant refactor of the router, which I'm probably not the best person to attempt due to my unfamiliarity with its intricacies.

johan-smits:
Is there a update on this?

PR #751, opened 2024-11-02
Router refresh

davidon-top:
closes: #335

### DIFF
--- a/packages/sycamore-router/src/router.rs
+++ b/packages/sycamore-router/src/router.rs
@@ -386,6 +386,8 @@ pub fn navigate_no_history(url: &str) {
         let pathname = pathname.get().unwrap_throw();
         let path = url.strip_prefix(&base_pathname()).unwrap_or(url);
         pathname.set(path.to_string());
+
+        window().scroll_to_with_x_and_y(0.0, 0.0);
     });
 }
 
@@ -403,6 +405,8 @@ pub fn refresh() {
         );
 
         pathname.get().unwrap_throw().update(|_| {});
+
+        window().scroll_to_with_x_and_y(0.0, 0.0);
     });
 }
 
--- a/packages/sycamore-router/src/router.rs
+++ b/packages/sycamore-router/src/router.rs
@@ -370,7 +370,9 @@ pub fn navigate_replace(url: &str) {
     });
 }
 
-/// Navigates to the specified `url`. Without doing anything with the history.
+/// Navigates to the specified `url` without touching the history API.
+///
+/// This means that the url will not be updated and will continue to show the previous value.
 ///
 /// # Panics
 /// This function will `panic!()` if a [`Router`] has not yet been created.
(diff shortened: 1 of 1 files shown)


packages/sycamore-router/src/router.rs
```diff
@@ -370,6 +370,35 @@ pub fn navigate_replace(url: &str) {
...
 
+/// Navigates to the specified `url`. Without doing anything with the history.
+///
+/// # Panics
+/// This function will `panic!()` if a [`Router`] has not yet been created.
+pub fn navigate_no_history(url: &str) {
+    PATHNAME.with(|pathname| {
+        assert!(
+            pathname.get().is_some(),
+            "navigate_no_hist can only be used with a Router"
```
lukechu10, line 381:
```suggestion
            "navigate_no_history can only be used with a Router"
```

packages/sycamore-router/src/router.rs
```diff
@@ -370,6 +370,35 @@ pub fn navigate_replace(url: &str) {
...
+        );
+
+        let pathname = pathname.get().unwrap_throw();
+        let path = url.strip_prefix(&base_pathname()).unwrap_or(url);
+        pathname.set(path.to_string());
+    });
```
lukechu10, line 387: I think we should also add this line here:
```rust
window().scroll_to_with_x_and_y(0.0, 0.0);
```
Although we probably want to make this configurable in the future.

packages/sycamore-router/src/router.rs
```diff
@@ -370,6 +370,35 @@ pub fn navigate_replace(url: &str) {
...
+}
+
+/// Refreshes the current page.
```
lukechu10, line 390:
```suggestion
/// Preform a "soft" refresh of the current page.
///
/// Unlike a "hard" refresh which corresponds to clicking on the refresh button, this simply forces a re-render of the view for the current page.
///
/// # Panic
/// This function will `panic!()` if a [`Router`] has not yet been created.
```

packages/sycamore-router/src/router.rs
```diff
@@ -370,6 +370,35 @@ pub fn navigate_replace(url: &str) {
...
+pub fn refresh() {
+    PATHNAME.with(|pathname| {
+        assert!(
+            pathname.get().is_some(),
+            "refresh can only be used with a Router"
+        );
+
+        pathname.get().unwrap_throw().update(|_| {});
```
lukechu10, line 398: Should also have scroll to top here, same comment as before.

packages/sycamore-router/src/router.rs
```diff
@@ -370,6 +370,35 @@ pub fn navigate_replace(url: &str) {
     });
 }
 
+/// Navigates to the specified `url`. Without doing anything with the history.
```
lukechu10, line 373:
```suggestion
/// Navigates to the specified `url` without touching the history API.
///
/// This means that the url will not be updated and will continue to show the previous value.
```

lukechu10 (approved):

Issue #23, opened 2021-03-08
JSX-like syntax macro


stoically:
Nice work on maple! Really like the VDOM-free reactivity approach and looking forward to see how this develops.

Wanted to leave a shameless plug here: in case you'd ever consider a JSX-like syntax for maple, then maybe my [syn-rsx](https://crates.io/crates/syn-rsx) crate might be of interest.

Feel free to close the issue if you consider this out of scope or an ad :D, and good luck with the library!

lukechu10:
Thanks for the feedback. Although I don't plan on supporting JSX anytime soon, I could see why some would prefer a JSX-like syntax (more familiar to React developers for one thing).

Personally, I am not to big a fan of JSX. It is a bit too verbose for my taste. I actually really like [Pug](https://pugjs.org/api/getting-started.html) but it relies on indentation for nesting which wouldn't work inside a proc-macro. In fact, the current `template!` macro is inspired form Pug syntax.

I'll leave this open as this could be a possibility in the future and I'll keep your crate in mind if I ever implement this :)

ufoscout:
Hi @lukechu10 
I really like the sycamore philosophy, anyway, I think that the syntax chosen for the template library reduces the userbase considerably. In fact, the standard HTML or HTML-like syntax is extremely familiar to everyone and it would definitely attract more people.
For example, when choosing a web framework for a project, in my team, we immediately discard every option that uses a custom template syntax because we don't want and we don't feel the need for learning something different.

lukechu10:
@ufoscout that's a good point. I think we could have something like [Marko's precise syntax](https://markojs.com/docs/concise/). The only thing I don't like about this is that it adds maintenance burden because now we need to maintain two codegens. Perhaps there could be a shared codegen "backend" and multiple codegen "frontends" (maybe even one with a builder-like syntax?).

pythoneer:
To add to the confusion: What i value most is easy integration into my development environment. I absolutely don't care if its JSX(-like) or any other syntax. Its easy to learn anyway and anything i came across (from other frameworks etc.) is manageable to use in a few minutes, so personally i don't see a problem here other than "taste" and in my case i can swallow that pill in exchange of a good experience. And what i mean with "good experience" is: can my development environment handle it to assist me with e.g. code completion, syntax checks, code formatting etc. 
[...]

PR #762, opened 2024-11-18
Move view.rs from sycamore-macro into sycamore-view-parser codegen.rs

davidon-top:
This would greatly simplify making 3rd party alternative view macros #23 only having to make the common ir rather then a custom solution that would work basically identically to the existing codegen solution.
Right now the view.rs was not accessible by other crates because the crate it is in is a proc-macro crate.
I know that technically the crate is named sycamore-view-**parser** and codegen isn't parsing but making a separate crate would be redundant in my opinion since the ir is also needed together with the codegen and i don't think its worth it renaming the crate since its already on crates.io.

Unrelated side note: I am working on and off on Fallible components trying to make it good. I just need a short change (e.g. making sycamore-rstml)

### DIFF
--- a/packages/sycamore-macro/src/lib.rs
+++ b/packages/sycamore-macro/src/lib.rs
@@ -10,7 +10,6 @@ use syn::{parse_macro_input, DeriveInput};
 mod component;
 mod inline_props;
 mod props;
-mod view;
 
 /// A macro for ergonomically creating complex UI complex layouts.
 ///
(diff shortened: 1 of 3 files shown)


lukechu10 (approved):
LGTM! I think moving the codegen to `sycamore-view-parser` makes sense as making the view parser separate was originally meant to allow creating alternative view syntax macros anyways.

PR #768, opened 2024-12-29
owned String required .into() or .to_string()

phocks:
Compiler was complaining about this. Apparently we need .into() or .to_string(). Feel free to change this to whatever is preferred.

In a previous Sycamore version, was this not required?


lukechu10 (approved):
Oops you’re right!

PR #780, opened 2025-01-15
Update resources-and-suspense.md

humb1t:
Replace `create_resource` with new `create_client_resource` name.


lukechu10:
Good catch thanks!

PR #783, opened 2025-01-23
Fix `unexpected_cfgs` in the crates and add a troubleshooting note

tguichaoua:
- Moves the `unexpected_cfgs` lint configration at the workspace level
- Inherits the workspace lints in all crates
- Add a notes about the "unexpected `cfg` condition name: `sycamore_force_ssr`" in the troubleshooting section of the docs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,4 +60,4 @@ sycamore-web = { path = "packages/sycamore-web", version = "0.9.1" }
 debug = true
 
 [workspace.lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(sycamore_force_ssr)'] }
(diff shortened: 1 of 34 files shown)


lukechu10:
Thanks for the PR and sorry for taking so long to get to this. I'm happy to merge this for now but I don't quite like this solution long term as it introduces extra friction to get started with Sycamore.

It seems like the main problem is that certain macros expand to `cfg(sycamore_force_ssr)` which then triggers the warning. Think behavior was introduced in https://github.com/rust-lang/rust/issues/132572

I think the best solution would be to not expand to `cfg(sycamore_force_ssr)` but rather to `true` or `false` depending on the flag. This way we do not get warnings in downstream crates.

PR #784, opened 2025-01-23
Fix lint warnings

tguichaoua:


lukechu10:
Yeah this is a bit annoying because the lint is telling us to use a new function but we can't because it's not supported on the MSRV. I guess the best solution would just be to update the MSRV...

tguichaoua:
I revert back the change for `is_none_or` and add a todo note.
I also used [`cagrgo-msrv`](https://github.com/foresterre/cargo-msrv) and the current MSRV of sycamore is `1.80`.
I don't know what is the MSRV policy for sycamore, but IMHO I think code quality isn't enough to justify to increase it.

lukechu10:
> I don't know what is the MSRV policy for sycamore, but IMHO I think code quality isn't enough to justify to increase it.

Yeah agreed. I think we'll update the MSRV in the next major release but I agree updating it in a minor release is not really worth it.

PR #785, opened 2025-02-01
Update todo-app.md

ZAZPRO:
Just followed introduction tutorial and faced an outdated part.

Web_sys window api is a feature and it is required for the local_storage access. And it returns an result so unwrap/expect is required.

[Docs](https://docs.rs/web-sys/0.3.77/web_sys/struct.Window.html#method.local_storage)

Also looks like KeyboardEvents are now part of sycamore, not web_sys.

### DIFF
--- a/docs/next/introduction/todo-app.md
+++ b/docs/next/introduction/todo-app.md
@@ -305,7 +305,7 @@ so that we can easily serialize `Signal`s.
 ```bash
 cargo add serde -F derive
 cargo add serde_json
-cargo add web-sys -F Storage
+cargo add web-sys -F "Storage,Window"
 cargo add sycamore -F serde
 ```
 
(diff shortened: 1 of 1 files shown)


lukechu10:
Sycamore has it’s own reexport of window which handles the unwrap already: https://docs.rs/sycamore/latest/sycamore/prelude/fn.window.html

Maybe we should make it more explicit that we import the prelude first?

docs/next/introduction/todo-app.md
```diff
@@ -355,8 +355,8 @@ Here is the complete code listing for the todo app.
 
 ```rust
 use serde::{Deserialize, Serialize};
-use sycamore::prelude::*;
-use web_sys::KeyboardEvent;
+use sycamore::{prelude::*, web::events::KeyboardEvent};
```
lukechu10, line 358: Very small thing but could you split this up into two use statements? This is how it is formatted in other places as well just for consistency.

ZAZPRO: I should have just copied and build the tutorial code before submitting this issue. Rust-analyzer was screaming at me that window function doesn't exist, that is why I searched for web_sys local storage requirements. I will close this pull request. Thanks for your time :)

PR #787, opened 2025-02-01
Fix missing deref in adding-state.md example

ZAZPRO:
I was referring to this part of the docs multiple times and each time I was puzzled, should I de-reference here or not.

Rust says I should.

Hope this will solve a confusion for future users.


lukechu10 (approved):
Thanks for catching this!

PR #796, opened 2025-03-29
fix reconcile_fragments bug

acheul:
The hashmap inside of reconcile_fragments() wrongly assigned values on its creation: it must be b_start+i, not i, to be correctly checked in following lines.


lukechu10:
Can you run cargo fmt? Otherwise looks great!

acheul:
Just done cargo fmt and cargo clippy! 👍

lukechu10 (approved):
Awesome thanks!

Issue #792, opened 2025-03-08
unexpected `cfg` condition name: `sycamore_force_ssr`


Tahinli:
**Describe the bug**
rust analyzer produces warning

```rust
warning: unexpected `cfg` condition name: `sycamore_force_ssr`
  --> src/main.rs:20:5
   |
20 |     console_log!("{}", "Tahinli");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: expected names are: `clippy`, `debug_assertions`, `doc`, `docsrs`, `doctest`, `feature`, `fmt_debug`, `miri`, `overflow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generalize_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `test`, `ub_checks`, `unix`, and `windows`
   = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
   = help: try referring to `$crate::is_ssr` crate for guidance on how handle this unexpected cfg
   = help: the macro `$crate::is_ssr` may come from an old version of the `sycamore_web` crate, try updating your dependency with `cargo update -p sycamore_web`
   = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
   = note: `#[warn(unexpected_cfgs)]` on by default
   = note: this warning originates in the macro `$crate::is_ssr` which comes from the expansion of the macro `console_log` (in Nightly builds, run with -Z macro-backtrace for more info)

warning: `rust_sycamore` (bin "rust_sycamore") generated 1 warning
```

**To Reproduce**
```rust
use sycamore::prelude::*;

fn main() {
    println!("Hello, world!");
    sycamore::render(App);
}

#[component]
fn App() -> View {
    let name = "Tahinli";
    view!(
        div {
            Hello(name=name)
        p {"Tahinli"}
    })
}

#[component(inline_props)]
fn Hello(name: &'static str) -> View {
    console_log!("{}", "Tahinli");
    view!(
        h1{"Hello," (name)}
    )
}
```

[...]

PR #798, opened 2025-07-01
Reimplement `is_ssr!` and `is_not_ssr!` macros to not trigger unknown cfg warning

lukechu10:
This is a better fix to #792 

Also deprecates `cfg_ssr_item!` and `cfg_not_ssr_item!` since they can just be replaced by `is_ssr!` and `is_not_ssr!`.


lukechu10:
Miri test failures seem to be irrelevant. Not sure what that is about...

PR #806, opened 2025-10-25
fix(802): nested batches to compose correctly

### LINKED ISSUE #802: Nested batches don't compose as expected
## Description

Based on my understanding of batch, I'd expect that when I nest batch calls:

- All updates from related signals are batched together, regardless of where they fall in the nesting hierarchy.
- Memos and effects only run at the end of the outermost batch.

As of Sycamore 0.9.1, however, it seems that:

- Updates in nested batches are batched separately.
- Memos and effects run at the end of each nested batch.

### Code

Run the following code by calling `cargo run`.

#### main.rs
```rust
use sycamore::prelude::*;

fn main() {
    let _ = create_root(|| {
        // create a state signal
        let state = create_signal("Initial");
        
        // monitor state updates
        let state_for_effect = state.clone();
        let mut update_count = 0;
        create_effect(move || {
            update_count += 1;
            console_log!("Update {update_count}: {}", state_for_effect.get());
        });
        
        // do some nested batched updates
        batch(|| {
            state.set("First in outer batch");
            batch(|| {
                state.set("First in inner batch");
                state.set("Last in inner batch");
            });
            state.set("Last in outer batch");
        });
    });
}
```

#### Cargo.toml
```toml
[package]
name = "nested-batches"
edition = "2021"

[dependencies]
sycamore = "0.9.1"

# turn off spurious warnings about the custom config that Sycamore uses
#
#   https://sycamore.dev/book/troubleshooting#unexpected-cfg-condition-name--sycamore-force-ssr
#
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ["cfg(sycamore_force_ssr)"] }
```

### Expected behavior

The console shows:
```
Update 1: Initial
Update 2: Last in outer batch
```

### Actual behavior

The console shows:
```
Update 1: Initial
Update 2: Last in inner batch
Update 3: Last in outer batch
```

## Environment

- Sycamore: 0.9.1
- Browser: none
- OS: Ubuntu 24.04

### Possible explanation

On Discord, @lukechu10 suggested [an explanation](https://discord.com/channels/820400041332179004/1397484079280292043/1397775690136424448) for the unintended behavior:

> Basically what batch does now is delay reactive state propagation until the end of the batch scope. This works by basically setting a global flag. However, when the batch scope ends, it does not check this flag to see if it is still set or not.

### Who needs nested batches?

If we expect nested batches to be redundant, why design for them at all? In the case where I hit this bug, the inner batch is part of a function, which sometimes gets called in isolation and other times gets called repeatedly within a larger batch.

Aditya-PS-05:
closes #802

### DIFF
--- a/packages/sycamore-reactive/src/root.rs
+++ b/packages/sycamore-reactive/src/root.rs
@@ -30,9 +30,10 @@ pub(crate) struct Root {
     pub nodes: RefCell<SlotMap<NodeId, ReactiveNode>>,
     /// A list of signals who need their values to be propagated after the batch is over.
     pub node_update_queue: RefCell<Vec<NodeId>>,
-    /// Whether we are currently batching signal updates. If this is true, we do not run
-    /// `effect_queue` and instead wait until the end of the batch.
-    pub batching: Cell<bool>,
+    /// The current batch depth. If greater than 0, don't run
+    /// `effect_queue` and instead wait until the end of the outermost batch.
+    /// This will make nested batches to compose correctly.
+    pub batch_depth: Cell<usize>,
 }
 
 thread_local! {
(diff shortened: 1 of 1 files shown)


lukechu10:
Sorry I honestly completely forgot about this. Is there a reason why you closed the PR?

Aditya-PS-05:
I closed this as there were no reviews.

PR #822, opened 2026-06-15
Warning when signal is accessed in component body

caibear:
Creates a warning when a signal is accessed inside a component's body which recommends that the programmer uses a closure. The message is based on the same warning in Leptos.

The implementation works and passes all the tests. The main reason it's still a draft is it creates panics, not warnings which would be a breaking change. It's unclear how sycamore-reactive can call e.g. `sycamore-web::console_warn!`.

### DIFF
--- a/packages/sycamore-core/src/component.rs
+++ b/packages/sycamore-core/src/component.rs
@@ -5,11 +5,11 @@ use std::fmt;
 use sycamore_reactive::*;
 
 /// Runs the given closure inside a new component scope. In other words, this does the following:
-/// * Create a new untracked scope (see [`untrack`]).
+/// * Create a new untracked scope (see [`untrack_in_component`]).
 /// * Call the closure `f` passed to this function.
 #[doc(hidden)]
 pub fn component_scope<T>(f: impl FnOnce() -> T) -> T {
-    untrack(f)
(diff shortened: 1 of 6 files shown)


caibear:
well so much for passing the tests :sweat_smile:, guess i forgot to enable all-features

lukechu10:
> It's unclear how sycamore-reactive can call e.g. `sycamore-web::console_warn!`

Hmm yeah I did not think about that 😅

Actually now that I think about it, perhaps a more elegant solution would be to simply use `tracked_scope(...)` inside of `component_scope` to see if any signals are accessed and display a warning/panic there instead. This can all happen inside the `sycamore-web` crate.

This way, we also don't "pollute" `sycamore-reactive` with things like component scopes.

The only downside with this approach is that the error message won't have the source location of where the signal was accessed, unless we add some callback hook to the DependencyTracker whenever a signal is accessed.

lukechu10:
So what I am thinking is adding a new `on_track: Option<Box<dyn Fn()>>` to `DependencyTracker` which will be called whenever a signal is tracked.

Then in `sycamore-web`, we can create a `DependencyTracker` when creating a component with a callback that will emit a warning when a signal is tracked.

Issue #670, opened 2024-04-20
Websocket example


UnlimitedCookies:
I suspect this is not yet implemented in sycamore, but an example might be quite useful.

lukechu10:
This would be pretty simple just using [`gloo-net`](https://docs.rs/gloo-net/latest/gloo_net/) but it might be good to have an example here. Another place this could go would be the https://github.com/lukechu10/sycamore-hooks although I haven't had time to work on that crate in a while.

Vectornaut:
I'm working on a WebSocket example, and I'd love to contribute it if I can get it working! Is `gloo-net` still the best crate to use?

I'll try to use the [WebSocket module](https://github.com/sycamore-community/sycamore-hooks/blob/master/src/websocket.rs) from `sycamore-hooks` as a guide, although I'm kind of confused about how it works.

lukechu10:
I was thinking it would probably be better for most people to just use the websocket module from sycamore-hooks instead of trying to roll their own wrapper. But perhaps we can have a minimal working example in this repository just using the raw web APIs. 

This should basically be the same way to do it as in JS but using the web-sys crate.

Vectornaut:
## Draft example

Here's a [first draft](https://codeberg.org/fareycircles/sycamore-websockets)! I'm currently using `gloo-net`, but could try to switch over to raw `web-sys` if you'd prefer.

The main barrier to moving this into the Sycamore repository is switching over to Sycamore's organization for examples that include server code. I just noticed that the [SSR streaming example](https://github.com/sycamore-rs/sycamore/tree/main/examples/ssr-streaming) does this, so maybe I could use that as a guide?

Here are some notes and questions.

- I made the shared state a [trigram](https://en.wikipedia.org/wiki/Bagua#Trigrams) so that it would be simple enough to fit into one byte, but complex enough to have more than one meaningful view on the client side. Another option I considered for the state was a 4 × 4 grid of black-and-white pixels, which fits exactly into two bytes. Would you prefer that? Or some other thing?
- Right now, the client and server are both very strict about validating messages. Every application state and user action has a single valid encoding as a WebSocket message, and the client and server will loudly reject any invalid messages. I could simplify the code by making the message format more flexible, with extraneous bits being silently ignored. How would you like me to balance modeling robust design and keeping the example code simple?
[...]

PR #824, opened 2026-07-03
Websockets example

Vectornaut:
## Summary

This branch addresses issue #670 by adding a basic example of a Sycamore web app that uses a WebSocket to communicate with a server. It includes both client and server code. To run the example, follow the instructions for "Examples that serve their client-side and server-side parts separately" that I've added to [`examples/README.md`](https://github.com/Vectornaut/sycamore/blob/websockets-example/examples/README.md#examples-that-serve-their-client-side-and-server-side-parts-separately).

The new example was initially developed as a standalone workspace in [this repository](https://codeberg.org/fareycircles/sycamore-websockets), and some discussion about it has already happened in the comments on issue #670.

## Organization

As of the initial pull request, the example is organized differently from other examples that use both client-side and server-side code. The organization may change during review.

### As of the initial pull request

The client and the server are laid out as separate binaries within the example crate, with common code in a top-level `lib.rs`. I set up `"client"` and `"server"` features that are required by the corresponding binaries and select each binary's dependencies. When you call `trunk serve`, the attributes of the `rust` link asset in `index.html` tell Trunk to enable the "client" feature and disable the default "server" feature. When you call `cargo run`, only the default "server" feature is enabled.

## Dependencies

The client depends on [`gloo-net`](https://docs.rs/gloo-net/latest/gloo_net/) in addition to Sycamore. The server is built with [`axum`](https://docs.rs/axum/latest/axum/), like in other examples that include server-side code.


Vectornaut:
Working on the formatting now! To gain familiarity with `rustfmt`, I'm looking through the changes carefully rather than just committing them after a quick once-over.

